PR #9, opened 2023-06-12
[16.0][MIG] edi_account_invoice_import

EvaSForgeFlow:
Correcting tests.

Depends on #5 and https://github.com/OCA/edi/pull/753

PR #8, opened 2023-06-12
[16.0][MIG] edi_ubl_oca

EvaSForgeFlow:
Correcting tests

Depends on #5

PR #10, opened 2023-06-12
[16.0][MIG] edi_account_invoice_import_ubl

EvaSForgeFlow:
Correcting tests

Depends on: #9 and #8 and https://github.com/OCA/edi/pull/740


LoisRForgeFlow:
Rebased. 

Commit history lost, pending to fix.

PR #7, opened 2023-06-09
[16.0][MIG] edi_endpoint_oca

JordiBForgeFlow:
Standard migration.

Todo:
* [ ] Functional review
* [ ] Review automated tests

PR #11, opened 2023-06-13
[MIG] edi_account_invoice_import_ubl_endpoint_example: Migration to 16.0

EvaSForgeFlow:
Correcting tests.

Depends on: #9 and #10 and  #7 and https://github.com/OCA/edi/pull/740


LoisRForgeFlow:
Rebased. 

Migration history lost, pending to fix.

PR #15, opened 2023-08-01
[16.0][MIG] edi_account_oca

OriolMForgeFlow:
Moved from edi repository. Standard migration.


OriolMForgeFlow:
HI @simahawk ,
I was wondering if you could take a look at this. The error I commented on in the previous PR, with the debug assets enabled, is as follows:

Screenshot from 2023-10-03 09-10-57

It appears when installing the 'edi_account_oca' module and you try to open the Invoicing app.

edi_account_oca/models/account_move.py
```diff
@@ -0,0 +1,37 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import fields, models
+
+
+class AccountMove(models.Model):
+    _name = "account.move"
+    _inherit = ["account.move", "edi.exchange.consumer.mixin"]
+
+    disable_edi_auto = fields.Boolean(
```
simahawk: field renamed and should override only the states

simahawk:
Here's the fix https://github.com/OCA/edi-framework/pull/23

edi_account_oca/models/account_move.py
OriolMForgeFlow: Fixed!

OriolMForgeFlow:
> Here's the fix #23

Thank you very much, @simahawk! 😄

This solution resolved the issue with rendering the view. However, a new problem has arisen: the view now appears somewhat disorganized. I also attempted to migrate edi_stock_oca and encountered the same issue.

Examples: 

Screenshot from 2023-10-04 09-27-32

Screenshot from 2023-10-04 09-45-28

simahawk:
Ok, it seems we have to find a better position where to place the buttons.

etobella:
@OriolMForgeFlow Can you test how it works with this?

https://github.com/OCA/edi-framework/pull/29

On my local repository worked fine

image

OriolMForgeFlow:
> @OriolMForgeFlow Can you test how it works with this?
> 
> #29
> 
> On my local repository worked fine
> 
> image

Worked correctly @etobella! When #29 is merged, this PR can also be reviewed.

etobella (approved):
Tested and all looked great

:+1:

PR #16, opened 2023-08-26
[16.0] edi_oca: port as is with all changes fix/imp from edi#14.0

simahawk:
The addon is not installable of course.


simahawk:
TODO: port https://github.com/OCA/edi/pull/790 when merged

simahawk:
https://github.com/OCA/edi/pull/790 included + added oca-port blacklist

PR #19, opened 2023-09-21
[16.0][ADD] edi_edifact_oca

QuocDuong1306:
This module provides a basic configuration for EDIFACT with the EDI framework.

Depends on:

- [x] OCA/partner-contact#1573
- [x] #20 
- [x] OCA/edi#813


edi_oca/__manifest__.py
```diff
@@ -10,7 +10,7 @@
     basic automation and views for handling EDI exchanges.
     """,
     "version": "15.0.1.5.0",
-    "website": "https://github.com/OCA/edi",
+    "website": "https://github.com/OCA/edi-framework",
```
simahawk: :thinking:

simahawk (changes requested):
As mentioned on https://github.com/OCA/edi/pull/813 the dependencies have to be fixed.
THIS module here is the on that should plug the party data module.

edi_edifact_oca/components/party_data.py
```diff
@@ -0,0 +1,12 @@
...
+
+from odoo.addons.component.core import Component
+
+
+class EDIPartyDataEDIFACT(Component):
+    """Party data provider specific to EDIFACT."""
+
+    _name = "edi.party.data.edifact"
+    _inherit = "edi.party.data"
+    _backend_type = "edifact"
```
simahawk, line 13: Can you add a TODO?  `TODO @simahawk: is a component per backend type really needed?`
I know you've taken it from edi_ubl_oca but we should try to remove special components when they are not needed IMO.

edi_edifact_oca/tests/test_party_data.py
```diff
@@ -0,0 +1,39 @@
...
+        cls.exc_record_edifact = cls.backend_edifact.create_record(
+            "edifact_out_test", {}
+        )
+        cls.partner = cls.env["res.partner"].create({"name": "EDIFACT Test"})
+        cls.exc_record_no_edifact = cls.backend_no_edifact.create_record(
+            "test_csv_output", {"model": cls.partner._name, "res_id": cls.partner.id}
+        )
+
+    def test_lookup_default(self):
+        provider = get_party_data_component(self.exc_record_no_edifact, self.partner)
```
simahawk: You are not testing at all your the code of this module... you are testing the base component... seems useless.

edi_edifact_oca/models/edifact.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, api, exceptions, models
+
+from odoo.addons.edi_party_data_oca.utils import get_party_data_component
+
+
+class BasePydifact(models.AbstractModel):
```
simahawk: Who should be the consumer of this method?

simahawk (changes requested):

edi_edifact_oca/models/edifact.py
QuocDuong1306: Yes, I think `get_party_data` was not used at all

yvaucher (approved):
LGTM

simahawk (approved):

PR #26, opened 2023-10-10
[16.0][MIG] edi_storage_oca: Migration to 16.0

QuocDuong1306:
- Migrate to v16
- Change the dependency from storage_backend to fs_storage


edi_storage_oca/components/base.py
```diff
@@ -0,0 +1,99 @@
...
+    def _remote_file_path(self, direction, state, filename):
+        """Return remote file path by direction and state for give filename.
+
+        :param direction: string stating direction of the exchange
+        :param state: string stating state of the exchange
+        :param filename: string for file name
+        :return: PurePath object
+        """
+        _logger.warning(
+            "Call of deprecated function `_remote_file_path`. "
```
simahawk: this should be trashed

edi_storage_oca/components/check.py
```diff
@@ -0,0 +1,92 @@
...
+                    {
+                        "edi_exchange_state": "output_sent_and_error",
+                        "exchange_error": pycompat.to_text(error_report),
+                    }
+                )
+                self.exchange_record._notify_error("process_ko")
+            return False
+        return True
+
+    # FIXME: this is not used ATM -> should be refactored
```
simahawk: this too

edi_storage_oca/components/listener.py
```diff
@@ -14,21 +14,16 @@ class EdiStorageListener(Component):
     def _move_file(self, storage, from_dir_str, to_dir_str, filename):
         from_dir = PurePath(from_dir_str)
         to_dir = PurePath(to_dir_str)
-        if filename not in storage.list_files(from_dir.as_posix()):
-            # The file might have been moved after a previous error.
+        # TODO:
+        # - storage.list_files now includes path in fs_storage, breaking change
+        # - we remove path
```
simahawk, line 19: not sure to understand this comment

edi_storage_oca/components/listener.py
```diff
@@ -14,21 +14,16 @@ class EdiStorageListener(Component):
...
+        files = storage.list_files(from_dir.as_posix())
+        files = [os.path.basename(f) for f in files]
+        if filename not in files:
             return False
-        self._add_after_commit_hook(
```
simahawk: Why do you remove this? This is needed to make the deletion happened only if the transaction is successful.

edi_storage_oca/tests/test_edi_backend_storage.py
```diff
@@ -39,7 +39,7 @@ def test_export_file_already_done(self):
         self.record.edi_exchange_state = "output_sent"
         mocked_paths = {self._file_fullpath("done"): self.fakepath}
         # TODO: test send only w/out cron (make sure check works)
-        self._test_run_cron(mocked_paths)
+        self._test_run_cron(mocked_paths, skip_sent=False)
```
simahawk, line 42: why do we have to add this change?

simahawk (changes requested):

edi_storage_oca/components/listener.py
QuocDuong1306: Sorry, i have a missing some where, I have undone it

edi_storage_oca/tests/test_edi_backend_storage.py
QuocDuong1306: Because in `tests` that have some case with the status `output_sent` so we use the flag `skip_sent` to ignore another case:
- `skip_sent` = True --> ignore records that were already sent
- `skip_sent` = False ---> send it

edi_storage_oca/components/check.py
```diff
@@ -0,0 +1,91 @@
...
+                    {
+                        "edi_exchange_state": "output_sent_and_error",
+                        "exchange_error": pycompat.to_text(error_report),
+                    }
+                )
+                self.exchange_record._notify_error("process_ko")
+            return False
+        return True
+
+    # into an incoming exchange.
```
simahawk: If you remove the FIXME comment and leave the method nobody will know that this must be removed.
IMO the whole method can be removed.

QuocDuong1306: Thank @simahawk, I have updated

edi_storage_oca/components/listener.py
QuocDuong1306: In `fs_storage`, the `list_files` method now returns filenames  with path. Ex: folderA/folderB/filename,..
About `backend_storage`, it just returns the filenames. So we make some changes on codes

simahawk: so... is it still a TODO or not?

edi_storage_oca/components/receive.py
```diff
@@ -0,0 +1,19 @@
...
+    _name = "edi.storage.component.receive"
+    _inherit = [
+        "edi.component.receive.mixin",
+        "edi.storage.component.mixin",
+    ]
+    _usage = "storage.receive"
+
+    def receive(self):
+        path = self._get_remote_file_path("pending")
+        filedata = self.storage.get(path.as_posix())
```
simahawk, line 19: `.get` as well as other methods are deprecated... can we at list add a TODO to clean this up? 
Can be added to ROADMAP file.

edi_storage_oca/tests/test_edi_backend_storage.py
simahawk: I know what it does :wink: 
I mean: why was not needed before and it is?

edi_storage_oca/components/listener.py
QuocDuong1306: Ah not TODO, just a note to understand this part code, I will remove TODO tag

simahawk (approved):

edi_storage_oca/models/edi_backend.py
```diff
@@ -0,0 +1,169 @@
...
+                if x.strip("/")
+            ]
+            return filenames
+
+        bits = [exchange_type.exchange_filename_pattern]
+        if exchange_type.exchange_file_ext:
+            bits.append(r"\." + exchange_type.exchange_file_ext)
+        pattern = "".join(bits)
+        # TODO: clean this up, .find_files is deprecated in fs_storage
+        full_paths = self.storage_id.find_files(pattern, full_input_dir_pending)
```
JordiMForgeFlow, line 164: Hi @QuocDuong1306 and @simahawk 

I am testing this PR and the method `find_files` (https://github.com/OCA/storage/blob/1a4d80faaab0ddf6dfc80d867472dd6ce4777c5f/fs_storage/models/fs_storage.py#L373) is not returning the full path from the root for me. Maybe you know what I may be missing.

I have an FTP where the home folder contains /import folder to import files. When the call to the method is done like `self.fs.ls(relative_path, detail=False)` being relative_path "/import", I just get the filename of the file inside the folder.

simahawk: I'll have a look soon

simahawk: @JordiMForgeFlow are you using `directory_path`? If yes, you might have inconsistent results. See https://github.com/OCA/storage/issues/311

JordiMForgeFlow: @simahawk I am not using directory_path. Nevertheless, I have been testing and I have been able to get it working fine by assuming that find_files returns only the filename, not the full path.

The problem is mainly that, if for example I have a file _test_file.xml_ in a folder called _import_, the find_files method returns _test_file.xml_. Then, when the line `return [p[pending_path_len:].strip("/") for p in full_paths]` is executed, the filename stored in the exchange record is left as _ile.xml_ (stripping the first characters as it assumes that it is actually the full path)

edi_storage_oca/components/base.py
```diff
@@ -0,0 +1,86 @@
...
+        :param filename: custom file name, exchange_record filename used by default
+        :return: remote file content as string
+        """
+        path = self._get_remote_file_path(state, filename=filename)
+        try:
+            # TODO: support match via pattern (eg: filename-prefix-*)
+            # otherwise is impossible to retrieve input files and acks
+            # (the date will never match)
+            # TODO: clean this up, .get is deprecated in fs_storage
+            return self.storage.get(path.as_posix(), binary=binary)
```
JordiMForgeFlow, line 69: @simahawk @QuocDuong1306  another issue that I am seeing is when sending a file. In the _send_ component, a check is done before sending to make sure a file is not sent twice. The check ends up in this line, doing a _storage.get_.

I suppose it has to do with the new fs storage, but doing the get for a file that does not exist raises a different error than FileNotFoundError. The error is:

`ValueError: <class 'ftplib.error_perm'>: "550 Can't open *****: No such file or directory"` (being ***** the filepath being accessed)

QuocDuong1306: I have created a PR at here: https://github.com/fsspec/filesystem_spec/pull/1494

john-herholz-dt:
What is the stage of this PR?

JordiMForgeFlow:
Hi @john-herholz-dt 

It works fine except for the issue I mentioned https://github.com/OCA/edi-framework/pull/26#discussion_r1429819870. I have been using a custom branch where I simply fixed what I commented and it has been working fine for some weeks already.

john-herholz-dt:
@JordiMForgeFlow ,
thanks for clarifying.

Why not add this fix to this branch?
At the moment the code coverage target isn't reached.

Will @simahawk merge it anyway or do we need to get the coverage?
Are you then continuing here or do you need help?

I need this module actually for 17.0 so I am waiting. =)
But thanks for the contribution so far, great work!

JordiMForgeFlow:
@john-herholz-dt 

The fix I added is basically changing `return [p[pending_path_len:].strip("/") for p in full_paths]` for `return [p.strip("/") for p in full_paths]`, as in my case (SFTP connection) I was not getting full paths. However, I am not sure if it has to do with my specific case or if it needs to be like this always.

simahawk:
@john-herholz-dt @JordiMForgeFlow @QuocDuong1306 If it's fine for you all I'd merge what we have here as it is and then we iterate again for fixes (whoever has the chance to do it 1st :wink:)

john-herholz-dt:
Super fine!

JordiMForgeFlow:
@simahawk LGTM! I will open a PR with the fix regarding the paths.

etobella:
@QuocDuong1306 why did you change storage_backend to fs_storage if there is no migration scripts there? If we migrate from  15 to 16, all the backends that are related to a storage will fail...

Actually, storage_backend module already exists on 16 and this module should replace the original module. But no migration scrips have been created besides a comment on an issue. 

@simahawk @JordiMForgeFlow @lmignon

simahawk:
@nilshamerlinck @QuocDuong1306 could you please have a look and evaluate how to get a script to migrate?
I would assume that `storage_backend` will still be installed in the old db and that we simply have to create the matching `fs.storage` records w/ their conf leaving the old `storage.backend` records untouched. 
The conversion part can be a util to be added to fs_storage but let's start small for now :wink:

nilshamerlinck:
@simahawk proposal of simple migration script pushed here : https://github.com/OCA/edi-framework/pull/95
note that sensitive parts of conf (host, port, credentials) is coming from the environment and will still need manual adaptation

PR #28, opened 2023-10-16
[16.0] [MIG] edi_exchange_template_oca: Migration to 16.0

etobella:
Migration to 16


etobella:
After some tests, this was the only way I found to fix it... somehow, odoo_test_helper is not restoring the original registry. This is not a problem on `at_install` tests, but might be a problem with `post_install` tests. Now it should be fixed  as all the affected tests have been transformed as `at_install`

Another option would be to add a second action, but this way seems cleaner to me

etobella:
@simahawk

simahawk:
> After some tests, this was the only way I found to fix it... somehow, odoo_test_helper is not restoring the original registry. This is not a problem on `at_install` tests, but might be a problem with `post_install` tests. Now it should be fixed as all the affected tests have been transformed as `at_install`
> 
> Another option would be to add a second action, but this way seems cleaner to me

You probably need this https://github.com/OCA/odoo-test-helper/pull/28

etobella:
Thanks @simahawk It works like a charm locally. Let me test on github actions :smile:

OriolMForgeFlow (approved):
LGTM! 👍🏼

pedrobaeza (approved):
/ocabot migration edi_exchange_template_oca
/ocabot merge nobump

PR #29, opened 2023-10-16
[16.0][FIX] edi_oca: Show buttons properly. Use OWL

etobella:


edi_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -115,7 +116,11 @@ def _get_eval_context(self):
     def get_view(self, view_id=None, view_type="form", **options):
         res = super().get_view(view_id, view_type, **options)
         if view_type == "form":
+            View = self.env["ir.ui.view"]
+            if view_id and res.get("base_model", self._name) != self._name:
+                View = View.with_context(base_model_name=res["base_model"])
             doc = etree.XML(res["arch"])
+            all_models = res["models"].copy()
```
simahawk, line 123: this backup is needed because...?

edi_oca/static/src/js/widget_edi.esm.js
```diff
@@ -0,0 +1,30 @@
...
+        ev.preventDefault();
+        ev.stopPropagation();
+        var typeId = rule.type.id;
+        const action = await this.orm.call(
+            this.props.record.resModel,
+            "edi_create_exchange_record",
+            [[this.props.record.resId], typeId],
+            {context: this.props.record.context}
+        );
+        this.env.services.action.doAction(action);
```
simahawk: this.action seems useless then...

edi_oca/static/src/xml/widget_edi.xml
```diff
@@ -1,22 +1,19 @@
...
+    <div t-name="edi_oca.EdiConfigurationWidget" owl="1">
+        <t t-foreach="props.value" t-as="rule_id" t-key="rule_index">
+            <t t-set="rule" t-value="props.value[rule_id]" />
+            <t t-if="rule.form">
+                <t t-set="btn" t-value="rule.form.btn" />
+                <button
+                    t-if="btn"
+                    t-on-click="(ev) => this.onClick(ev, rule)"
+                    t-att-title="btn.tooltip"
+                    t-attf-class="o_edi_create_exchange_record btn #{btn.class != null ? btn.class : 'btn-sm btn-secondary'}"
```
simahawk:
```suggestion
                    t-attf-class="o_edi_action btn #{btn.class != null ? btn.class : 'btn-sm btn-secondary'}"
```

edi_oca/templates/exchange_mixin_buttons.xml
```diff
@@ -3,7 +3,7 @@
     <template id="edi_exchange_consumer_mixin_buttons">
         <div
             t-att-groups="group"
-            class="alert alert-warning"
+            class="o_mg_alert alert-warning"
```
simahawk: this looks wrong.  The `o_mg_alert` is only for mail_group alerts. Plus, even if you keep it the `alert` class should stay.

simahawk (changes requested):
Thanks for taking care of this!

Judging from what I see on https://github.com/OCA/edi-framework/pull/15 I think we need a bit of padding (probably due to the missing alert class).

edi_oca/templates/exchange_mixin_buttons.xml
etobella: Yes, you are right. I had a confusion here....
image

edi_oca/models/edi_exchange_consumer_mixin.py
etobella: Because I forgot to review if new models are added on the new process_view. Now it should work properly

etobella:
@simahawk All comments attended

edi_oca/models/edi_exchange_consumer_mixin.py
simahawk: a comment would help our future selves reminding why :wink:

edi_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -127,18 +132,18 @@ def get_view(self, view_id=None, view_type="form", **options):
...
-            new_models.update(res["models"])
-            res["models"] = new_models
+                new_node = etree.fromstring(str_element)
+                new_arch, new_models = View.postprocess_and_fields(new_node, self._name)
+                for model in new_models:
+                    if model in all_models:
+                        continue
+                    if model not in res["models"]:
+                        all_models[model] = new_models[model]
+                    else:
```
simahawk: this `else` is superflous

simahawk: Do we really need this handling of the models?
I expect `models` to contain already the current model as you called super...
Just trying to understand how this works, sorry if I miss any piece of the puzzle :sweat_smile:

OriolMForgeFlow (approved):
Code review and tested in our environment. Works properly! 😄

edi_oca/models/edi_exchange_consumer_mixin.py
etobella: Yes, but we need to ensure that no new models are added (for example on child records). For this reason we need to extend. The same logic is applied in base_tier_validation.
 I know that that should happen, but it is safer to keep it.

edi_oca/models/edi_exchange_consumer_mixin.py
simahawk: ping

simahawk (approved):
minor remark on a missing comment, otherwise LG

PR #31, opened 2023-10-18
[16.0][MIG] edi_sale_oca: Migration to 16.0

QuocDuong1306:
Migrate from:
- https://github.com/OCA/edi/pull/831

Depends on:

- https://github.com/OCA/edi/pull/816


simahawk:
> In [OCA/edi#759](https://github.com/OCA/edi/pull/759), should we need to move `listeners.py` in `edi_sale_ubl_oca` to `edi_sale_oca` for common?
> Because in the future, maybe we need more modules like `edi_sale_ubl_oca` (ex: `edi_sale_edifact_oca`,...)

I'd do it only if we think that setting the state is needed in all the other cases is needed.
Otherwise, what would you move? An empty listener for create and write? 

What is your POV on this?

QuocDuong1306:
Yes @simahawk, you're right. We need to move when all standards need setting the state. So please ignore it and could you review this module? Thank a lot

yvaucher:
Please squash Simone's commits

[fixup! Add edi_sale_oca](https://github.com/OCA/edi-framework/pull/31/commits/c782edc3a45edb04c2ebac19bd794f76c8b8f7b6)

[fixup! fixup! Add edi_sale_oca](https://github.com/OCA/edi-framework/pull/31/commits/2afa60eb2d3806e7cceec79c64b34e741d3a56bd)


into

[Add edi_sale_oca](https://github.com/OCA/edi-framework/pull/31/commits/d88a1ffb4c24017467ed7a17fb32e7b24d4f4629)

test-requirements.txt
```diff
@@ -1 +1,2 @@
 odoo_test_helper
+odoo-addon-sale_order_import @ git+https://github.com/OCA/edi.git@refs/pull/816/head#subdirectory=setup/sale_order_import
```
yvaucher: Can be removed as https://github.com/OCA/edi/pull/816 was just merged

yvaucher (commented):
Squash to do and test-requirements.txt to update

Otherwise LGTM

QuocDuong1306:
Hi @yvaucher , I updated. When squashing `fixup! fixup! Add edi_sale_oca` into `Add edi_sale_oca`, it was conflicted. So I squash `fixup! fixup! Add edi_sale_oca` into `edi_sale: clean file` and change the tile commit to `edi_sale: clean file and drop *type_auto dependency`. Thanks

edi_sale_oca/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,23 @@
...
+You can decide if the order should be confirmed by exchange type.
+
+On your exchange type, go to advanced settings and add the following::
+
+    [...]
+    components:
+        process:
+            usage: input.process.sale.order
+    [...]
+    sale_order:
```
simahawk: if we do the change mentioned above... this has to be fixed

simahawk:
> Hi @yvaucher , I updated. When squashing `fixup! fixup! Add edi_sale_oca` into `Add edi_sale_oca`, it was conflicted. So I squash `fixup! fixup! Add edi_sale_oca` into `edi_sale: clean file` and change the tile commit to `edi_sale: clean file and drop *type_auto dependency`. Thanks

In theory it should have been squashed on the original PR... :wink:

QuocDuong1306:
I updated

edi_sale_oca/components/process.py
```diff
@@ -63,27 +63,17 @@ def _setup_wizard(self):
...
-        wiz.price_source = self._get_default_price_source()
         return wiz
 
-    @api.model
-    def _get_default_price_source(self):
-        return self.settings.get("price_source", "pricelist")
-
-    def _order_should_be_confirmed(self):
-        return self.settings.get("confirm_order", False)
-
```
QuocDuong1306: We can use `default_*` from EDI context, so I removed them

edi_sale_oca/components/process.py
```diff
@@ -0,0 +1,89 @@
...
+
+    _name = "edi.input.sale.order.process"
+    _inherit = "edi.component.input.mixin"
+    _usage = "input.process.sale.order"
+
+    def __init__(self, work_context):
+        super().__init__(work_context)
+        self.settings = {}
+        # Suppor legacy key `self.type_settings`
+        for key in ("sale_order", "sale_order_import"):
```
simahawk: IMO we can remove all these special keys. Including `wiz_ctx`.

If fixed a long time ago the propagation of `env_ctx` down the edi stack.
This means that, in theory, we can pass all configurations via `env_ctx`.

IMO this has a bunch of benefits:

1. contextualize the conf inside the specific component's settings
2. eliminate superfluous code

We could even keep the `wiz_ctx` inside the `env_ctx` but I wonder if it makes sense. 
We could get rid of line 56 and simply have something like:

```
components:
    process:
        usage: input.process.sale.order
        env_ctx:
            # values for the wizard
            default_confirm_order: true
            default_price_source: order
            # custom keys, whatever you need
            random_one: true
```

I'm sorry because I should have written this stuff in the roadmap but is better to do it now before we start using it otherwise we'll have to support the old style conf for another bunch of versions.
If you can go ahead w/ this, pls do it in its own commit to be able to cherry-pick it.

QuocDuong1306: I updated, could you have a look?

edi_sale_oca/components/process.py
```diff
@@ -0,0 +1,86 @@
...
+
+    _name = "edi.input.sale.order.process"
+    _inherit = "edi.component.input.mixin"
+    _usage = "input.process.sale.order"
+
+    def __init__(self, work_context):
+        super().__init__(work_context)
+        self.settings = {}
+        process_ctx = self.type_settings.get("process", {})
+        self.settings = process_ctx.get("env_ctx", {})
```
simahawk: You don't need this. The `env_ctx` is already propagated to the collection's context and to the record context.
The one of the collection is the one you have on self.env since the env comes from the collection.
See below.

edi_sale_oca/components/process.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def _setup_wizard(self):
+        """Init a `sale.order.import` instance for current record."""
+        ctx = self.settings
+        # Set the right EDI origin on both order and lines
+        edi_defaults = {"origin_exchange_record_id": self.exchange_record.id}
+        ctx["sale_order_import__default_vals"] = dict(
+            order=edi_defaults, lines=edi_defaults
+        )
+        wiz = self.env["sale.order.import"].with_context(**ctx).sudo().create({})
```
simahawk:
```suggestion
        # Set the right EDI origin on both order and lines
        edi_defaults = {"origin_exchange_record_id": self.exchange_record.id}
        addtional_ctx = dict(
            sale_order_import__default_vals=dict(
                order=edi_defaults, lines=edi_defaults
            )
        )
        wiz = self.env["sale.order.import"].with_context(**addtional_ctx).sudo().create({})
```

edi_sale_oca/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,26 @@
+TODO
```
simahawk:
```suggestion
```

edi_sale_oca/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,26 @@
...
+    components:
+        process:
+            usage: input.process.sale.order
+        env_ctx:
+                # Values for the wizard
+                default_confirm_order: true
+                default_price_source: order
+                # Custom keys, whatever you need
+                random_one: true
+
```
simahawk, line 22:
```suggestion

Note that `env_ctx` will propagate all keys to the whole env so you can use it for any kind of context related configuration. In the case of the sale order import wizard here we are just passing defaults as we could do in odoo standard.
```

simahawk (changes requested):

edi_sale_oca/components/process.py
simahawk: See https://github.com/OCA/edi-framework/blob/16.0/edi_oca/models/edi_backend.py#L61C48-L61C48

simahawk (approved):

simahawk:
ah well, the OCA infra is down now... won't get any merge

nilshamerlinck:
@simahawk looks like Alexandre fixed the server issue, so the gitbot should be back, would you mind issuing the command again?

PR #33, opened 2023-10-24
[16.0][ADD] edi_exchange_record_company

OriolMForgeFlow:


OriolMForgeFlow:
Better to add it in edi_oca base module: https://github.com/OCA/edi-framework/pull/34

PR #34, opened 2023-10-25
[16.0][IMP] edi_oca: add company_id field with multi-company rule

OriolMForgeFlow:


etobella (approved):
@simahawk Shall we use the default value by default?

edi_oca/models/edi_exchange_record.py
```diff
@@ -108,6 +108,7 @@ class EDIExchangeRecord(models.Model):
         compute="_compute_retryable",
         help="The record state can be rolled back manually in case of failure.",
     )
+    company_id = fields.Many2one("res.company", string="Company", readonly=True)
```
simahawk: why readonly?

simahawk:
IMO

> @simahawk Shall we use the default value by default?

I don't think so. The rule is relaxed enough to work by default w/o company.
Am I missing any drawback?

etobella:
It was just a question, it is good for me.  Let's wait for the readonly and we can move on with it :smile:

edi_oca/models/edi_exchange_record.py
OriolMForgeFlow: I thought it was good idea to prevent users to accidentally or intentionally change the company associated with a record. I hadn't thought of any situation where someone would want to do that. But, nvm, I have changed it.

simahawk: @OriolMForgeFlow too late, it has been merged... :smile: 
I get your point, feel free to propose a change.
Also, regarding the default: feel free to propose a change as well :wink:

PR #42, opened 2023-11-22
[16.0][POC] edi_oca: relate multiple records with a single exchange record

JordiMForgeFlow:
In our projects we have faced different scenarios where we have needed to link more than one Odoo record to an exchange record. For example, in a warehouse integration, you could be exchanging a Product Masterdata file including multiple products, and you want them all to be linked to the exchange record.

Currently the EDI framework only allows to link one Odoo record to an exchange record. Using the current code, a solution could be to use the ability to link exchange records in a parent/child relation. However, in the use case presented you will only sent one file, so it is a bit weird to have child exchange records that will simply do nothing.

My goal is to try to find a solution that fits the current code and allows covering these type of use cases. As a generic approach, I am proposing this POC where we could simply add a model to relate multiple records (being from different models) with a single exchange record. Obviously, if this is agreed as a good solution, we would then work in this PR to implement it properly.

Looking forward to get your opinions!

CC @etobella @simahawk @JaumeBforgeFlow 

- [x] Add new model to relate odoo records with a exchange record
- [x] Allow navigating to the related odoo record from the exchange record
- [ ] Always use related records (main record is computed as the first record of the related records)

### DIFF
--- a/edi_oca/models/edi_exchange_record.py
+++ b/edi_oca/models/edi_exchange_record.py
@@ -7,7 +7,7 @@ import base64
 import logging
 from collections import defaultdict
 
-from odoo import _, api, exceptions, fields, models
+from odoo import Command, _, api, exceptions, fields, models
 
 _logger = logging.getLogger(__name__)
 
@@ -35,7 +35,14 @@ class EDIExchangeRecord(models.Model):
     )
     direction = fields.Selection(related="type_id.direction")
     backend_id = fields.Many2one(comodel_name="edi.backend", required=True)
-    model = fields.Char(index=True, required=False, readonly=True)
+    model = fields.Char(
+        index=True,
+        required=False,
+        readonly=True,
+        compute="_compute_record",
+        inverse="_inverse_record",
+        store=True,
+    )
     res_id = fields.Many2oneReference(
         string="Record",
         index=True,
@@ -43,6 +50,9 @@ class EDIExchangeRecord(models.Model):
         readonly=True,
         model_field="model",
         copy=False,
+        compute="_compute_record",
+        inverse="_inverse_record",
+        store=True,
     )
     related_record_exists = fields.Boolean(compute="_compute_related_record_exists")
     related_name = fields.Char(compute="_compute_related_name", compute_sudo=True)
@@ -87,9 +97,14 @@ class EDIExchangeRecord(models.Model):
         help="Original exchange which originated this record",
     )
     related_exchange_ids = fields.One2many(
-        string="Related records",
+        string="Related exchanges",
         comodel_name="edi.exchange.record",
         inverse_name="parent_id",
+    )
+    related_record_ids = fields.One2many(
+        string="Related records",
+        comodel_name="edi.exchange.related.record",
+        inverse_name="exchange_record_id",
     )
     ack_expected = fields.Boolean(compute="_compute_ack_expected")
     # TODO: shall we add a constrain on the direction?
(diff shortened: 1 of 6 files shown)


edi_oca/models/edi_exchange_record.py
```diff
@@ -91,6 +95,11 @@ class EDIExchangeRecord(models.Model):
         comodel_name="edi.exchange.record",
         inverse_name="parent_id",
     )
+    secondary_record_ids = fields.One2many(
+        string="Secondary records",
+        comodel_name="edi.exchange.record.secondary.record",
+        inverse_name="exchange_record_id",
+    )
```
```suggestion
    related_record_ids = fields.One2many(
        string="Related records",
        comodel_name="edi.exchange.related.record"
        inverse_name="exchange_record_id",
    )
```
And then:
* always use the relation instead of the direct reference
* make model and res_id computed on the 1st record
* adapt related record methods

edi_oca/models/edi_exchange_record.py
```diff
@@ -149,6 +158,19 @@ def _constrain_edi_exchange_state(self):
                     _("Exchange state must respect direction!")
                 )
 
+    @api.constrains("secondary_record_ids")
+    def _check_secondary_record_models(self):
+        """
+        Check that all secondary records have the same model.
```
simahawk: Why? IMO we should support different kind of records.

edi_oca/models/edi_exchange_record.py
```diff
@@ -172,6 +194,11 @@ def _compute_related_record_exists(self):
         for rec in self:
             rec.related_record_exists = bool(rec.record)
 
+    @api.depends("secondary_record_ids")
+    def _compute_related_secondary_records_exists(self):
+        for rec in self:
+            rec.related_secondary_records_exists = any(rec.secondary_record_ids)
+
```
```suggestion
```

edi_oca/models/edi_exchange_record.py
```diff
@@ -351,6 +378,39 @@ def action_open_related_record(self):
             return {}
         return self.record.get_formview_action()
 
+    def action_open_related_secondary_records(self):
```
simahawk: To jump to all related records we have 2 options:
1. go to the list of each record and click on each specific action
2. generate 1 btn per each model automatically in the form

edi_oca/models/edi_exchange_record_secondary_record.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2023 ForgeFlow S.L.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import fields, models
+
+
+class EDIExchangeRecordSecondaryRecord(models.Model):
```
simahawk: see comments above

edi_oca/models/edi_exchange_record_secondary_record.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    Model linking an Odoo record with an exchange record.
+    """
+
+    _name = "edi.exchange.record.secondary.record"
+    _description = "EDI exchange Record Secondary Record"
```
simahawk: order by id so that the 1st record is always the 1st one (to compute the main record mostly)

simahawk (changes requested):
@JordiMForgeFlow thanks for the proposal. I hope my counter proposal is clear :)

edi_oca/models/edi_exchange_record.py
JordiMForgeFlow: True, we also though about it in the beginning but were not sure how to handle then the action to open the related records. However, I can see you suggested 2 options to handle this so then we could allow different models :)

edi_oca/models/edi_exchange_related_record.py
```diff
@@ -0,0 +1,40 @@
...
+        string="Record",
+        index=True,
+        required=True,
+        readonly=True,
+        model_field="model",
+        copy=False,
+    )
+
+    @property
+    def record(self):
```
simahawk, line 41: I would make this a Many2oneReference field and leave res_id as an integer only

PR #47, opened 2023-12-21
[17.0][MIG] edi_oca: Migration to 17.0

john-herholz-dt:
A lot of views needed to be adjusted.

Furthermore I changed the _search method on edi_exchange_record, as the count param no longer works.
Not sure about the change as I don't know what the override of the _search method was suppossed to do.

Edit: @nguyenminhchien helped on the _search method.

PR #52, opened 2024-01-09
[16.0][IMP] edi_oca: add partner form page

QuocDuong1306:
Fix the TODO at: https://github.com/OCA/edi-framework/blob/16.0/edi_sale_oca/views/res_partner.xml#L13

### DIFF
--- a/edi_account_oca/views/res_partner.xml
+++ b/edi_account_oca/views/res_partner.xml
@@ -7,6 +7,7 @@
         <field name="model">res.partner</field>
         <field name="inherit_id" ref="account.view_partner_property_form" />
         <field name="arch" type="xml">
+            <!-- TODO: Move this conf inside `edi` page from edi_oca. -->
             <group name="accounting_entries" position="after">
                 <group name="edi_configuration" string="EDI Configuration" />
             </group>
--- a/edi_oca/views/res_partner.xml
+++ b/edi_oca/views/res_partner.xml
@@ -9,11 +9,6 @@
         <field name="arch" type="xml">
             <xpath expr="//page[last()]" position="after">
                 <page name="edi" string="EDI">
-                    <!--
-                        TODO: IMO this page must be provided by edi_oca
-                        and all the modules should hook here instead of adding their own blocks
-                        here and there (see edi_account_oca, edi_stock_oca, etc)
-                    -->
                     <group name="edi_main" />
                 </page>
             </xpath>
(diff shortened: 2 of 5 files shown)


edi_oca/views/res_partner.xml
```diff
@@ -1,6 +1,4 @@
 <?xml version="1.0" encoding="utf-8" ?>
-<!-- Copyright 2022 Camptocamp SA
```
etobella: Don't remove the copyright

etobella (changes requested):
I do not agree with this change on an already installed version. This could affect other modules accross other repos (inside OCA or not). At the end, we need to prevent this kind of changes on Mature modules (I know, it says Beta, but after 4 versions, we can talk of a Mature module....). I would recommend to apply this change on  thenew release for 17 #47

simahawk:
I think we can find a solution here: do not touch the other modules. They are adding a group in a specific section, not a page. Just move the view from edi_sale to core.
I would also add a TODO in those modules to move their conf to the same page in their own group.

simahawk:
As you are there, please rewrite the commit as 
```
[IMP] edi_oca: add partner form page
```

QuocDuong1306:
Thank @simahawk @etobella , I updated

edi_account_oca/views/res_partner.xml
```diff
@@ -7,6 +7,7 @@
         <field name="model">res.partner</field>
         <field name="inherit_id" ref="account.view_partner_property_form" />
         <field name="arch" type="xml">
+            <!-- TODO: Move this conf to the EDI page in edi_oca. -->
```
simahawk:
```suggestion
            <!-- TODO: Move this conf inside `edi` page from edi_oca. -->
```

edi_stock_oca/views/res_partner.xml
```diff
@@ -7,6 +7,7 @@
         <field name="model">res.partner</field>
         <field name="inherit_id" ref="stock.view_partner_stock_form" />
         <field name="arch" type="xml">
+            <!-- TODO: Move this conf to the EDI page in edi_oca. -->
```
simahawk:
```suggestion
            <!-- TODO: Move this conf inside `edi` page from edi_oca. -->
```

simahawk (approved):
Minor remarks, otherwise ok for me.
@etobella better?

etobella (approved):

etobella (approved):
/ocabot merge minor

PR #46, opened 2023-12-18
 [16.0][MIG] edi_endpoint_oca

simahawk:
Replaces  #7, restore the origin field, fixes pre-commit.

PR #55, opened 2024-01-15
[FIX] edi_oca: tests using FakeModelLoader should not be post_install

nilshamerlinck:
This PR will fix the issues faced in https://github.com/OCA/edi-framework/pull/46

The rationale behind this change is explained here: https://github.com/OCA/odoo-test-helper/pull/30


simahawk:
/ocabot merge patch

NOTE: to be backported

PR #58, opened 2024-01-16
[16.0][IMP] edi_oca: add encoding management for exchange type

JordiMForgeFlow:


JordiMForgeFlow:
@simahawk @etobella could I get some reviews here? :)

edi_oca/models/edi_exchange_type.py
```diff
@@ -151,6 +151,39 @@ class EDIExchangeType(models.Model):
         "you should define a sequence in this field to show "
         "the sequence in your filename",
     )
+    # https://docs.python.org/3/library/codecs.html#standard-encodings
+    encoding_out = fields.Char(
```
etobella: Why do you need out and in? An edi.exchange.type can only be input or output, but not both :thinking: 

With one might be enough, isn't it?

JordiMForgeFlow: @etobella true, for the encoding I can keep one field, although for the error handler there is one option that is only available for encoding, so maybe we can keep it separated and only show the relevant field depending on the direction.

edi_oca/models/edi_exchange_type.py
```diff
@@ -151,6 +151,36 @@ class EDIExchangeType(models.Model):
         "you should define a sequence in this field to show "
         "the sequence in your filename",
     )
+    # https://docs.python.org/3/library/codecs.html#standard-encodings
+    encoding = fields.Char(
+        help="Encoding to be applied to generate/process the exchanged file.\n"
+        "Example: UTF-8, Windows-1252, ASCII...(default is always 'UTF-8')",
+    )
+    # https://docs.python.org/3/library/codecs.html#codec-base-classes
+    encoding_out_error_handler = fields.Selection(
```
etobella, line 160: the same for the other field

edi_oca/models/edi_exchange_type.py
etobella: Ok, makes sense to me

etobella (approved):

edi_oca/tests/test_exchange_type_encoding.py
```diff
@@ -0,0 +1,84 @@
+# Copyright 2024 ForgeFlow S.L. (https://www.forgeflow.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+import base64
+
+import chardet
```
simahawk, line 5: Do we really need this test dep?
Can't we have some pre-defined txt w/ specific encoding?

edi_oca/models/edi_exchange_type.py
```diff
@@ -151,6 +151,36 @@ class EDIExchangeType(models.Model):
...
+        string="Encoding Error Handler",
+        selection=[
+            ("strict", "Raise Error"),
+            ("ignore", "Ignore"),
+            ("replace", "Replace with Replacement Marker"),
+            ("backslashreplace", "Replace with Backslashed Escape Sequences"),
+            ("surrogateescape", "Replace Byte with Individual Surrogate Code"),
+            ("xmlcharrefreplace", "Replace with XML/HTML Numeric Character Reference"),
+        ],
+        help="Handling of encoding errors (default is always 'Raise Error').",
```
simahawk:
```suggestion
        help="Handling of encoding errors on generate (default is always 'Raise Error').",
```

edi_oca/models/edi_exchange_type.py
```diff
@@ -151,6 +151,36 @@ class EDIExchangeType(models.Model):
...
+    encoding_in_error_handler = fields.Selection(
+        string="Decoding Error Handler",
+        selection=[
+            ("strict", "Raise Error"),
+            ("ignore", "Ignore"),
+            ("replace", "Replace with Replacement Marker"),
+            ("backslashreplace", "Replace with Backslashed Escape Sequences"),
+            ("surrogateescape", "Replace Byte with Individual Surrogate Code"),
+        ],
+        help="Handling of decoding errors (default is always 'Raise Error').",
```
simahawk:
```suggestion
        help="Handling of decoding errors on process (default is always 'Raise Error').",
```

simahawk (changes requested):
~blocking point just for the test dep, otherwise~ LG

simahawk:
> blocking point just for the test dep, otherwise LG

bah nevermind... it's an odoo dep :sweat:

simahawk:
just a minor remark to make the help txt more clear. Does it help?

JordiMForgeFlow:
@simahawk changes applied, thank you! :)

simahawk (approved):

PR #61, opened 2024-01-17
[16.0][IMP] edi_oca: introduce a deduplication mechanism at the sending step

QuocDuong1306:
Rebased on:
- #63 

Example problematic scenario:
- User creates a new customer
- User goes through n rounds of (Edits, Saves)
- This generates n edi.exchange.record
- Given that we always send a full representation of the customer to WAMAS (and not incremental changes), we only need the last exchange record to be sent
- Sending more than the last echange record can create trouble:
 - if for some reason, WAMAS does not process the records in their filename order, it could save an outdated representation of the customer
 - too many records to process might clog queues and/or create delays/high load
 
Solution:
- Introduce a deduplication mechanism in edi_oca at the sending step

Pros:
- If needed for debugging , we can keep track of all the superseded records
- If last change on Customer ends up with a failure of action_exchange_generate, the previous representation will be sent
- The Output Sync cron controls the buffering window: if run every 5 minutes, we'll have at max one WAMAS file every 5 minutes (per customer)

Cons:
- We still generate a lot of queue jobs (generate, send),

### DIFF
--- a/edi_oca/data/cron.xml
+++ b/edi_oca/data/cron.xml
@@ -33,4 +33,20 @@
         <field name="state">code</field>
         <field name="code">model.search([])._cron_check_input_exchange_sync()</field>
     </record>
+    <record
+        id="cron_edi_backend_delete_obsolete_records"
+        model="ir.cron"
+        forcecreate="True"
+    >
+    <field name="name">EDI exchange delete obsolete records</field>
+    <field name="active" eval="True" />
+    <field name="user_id" ref="base.user_root" />
+    <field name="interval_number">1</field>
+    <field name="interval_type">days</field>
+    <field name="numbercall">-1</field>
+    <field name="doall" eval="False" />
+    <field name="model_id" ref="edi_oca.model_edi_backend" />
+    <field name="state">code</field>
+    <field name="code">model.search([])._cron_delete_obsolete_records()</field>
+</record>
 </odoo>
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -284,6 +284,10 @@ class EDIBackend(models.Model):
         check = self._output_check_send(exchange_record)
         if not check:
             return "Nothing to do. Likely already sent."
+        # In case obsolete: skip sending
+        is_obsolete = self._check_obsolete(exchange_record)
+        if is_obsolete:
+            return "Nothing to do. Likely obsolete."
         state = exchange_record.edi_exchange_state
         error = False
         message = None
@@ -670,3 +674,43 @@ class EDIBackend(models.Model):
             if raise_if_not:
                 raise
             return False
+
+    def _check_obsolete(self, exchange_record):
+        res = False
+        if exchange_record.is_obsolete:
+            exchange_record.write({"edi_exchange_state": "obsolete"})
+            res = True
+        return res
+
+    def _cron_delete_obsolete_records(self, **kw):
+        for backend in self:
+            backend._delete_obsolete_records(**kw)
+
+    def _delete_obsolete_records(self, record_ids=None, **kw):
+        """Cleanup obsolete records.
+
+        Go through types with `delete_obsolete_records` flag on
+        and delete their obsolete records if any.
+        """
+        obsolete_records = self.exchange_record_model.search(
+            self._obsolete_records_domain(record_ids=record_ids)
+        )
+        _logger.info(
+            "EDI Exchange delete records: found %d obsolete records to delete.",
+            len(obsolete_records),
+        )
+        if obsolete_records:
+            obsolete_records.unlink()
+
+    def _obsolete_records_domain(self, record_ids=None):
+        """Domain for obsolete records need to delete."""
+        domain = [
+            ("backend_id", "=", self.id),
+            ("type_id.direction", "=", "output"),
+            ("type_id.delete_obsolete_records", "=", True),
+            ("edi_exchange_state", "=", "obsolete"),
+            ("exchange_file", "!=", False),
+        ]
+        if record_ids:
+            domain.append(("id", "in", record_ids))
+        return domain
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -706,10 +706,8 @@ class EDIBackend(models.Model):
         """Domain for obsolete records need to delete."""
         domain = [
             ("backend_id", "=", self.id),
-            ("type_id.direction", "=", "output"),
             ("type_id.delete_obsolete_records", "=", True),
             ("edi_exchange_state", "=", "obsolete"),
-            ("exchange_file", "!=", False),
         ]
         if record_ids:
             domain.append(("id", "in", record_ids))
--- a/edi_oca/models/edi_exchange_record.py
+++ b/edi_oca/models/edi_exchange_record.py
@@ -70,6 +70,7 @@ class EDIExchangeRecord(models.Model):
         selection=[
             # Common states
             ("new", "New"),
+            ("obsolete", "Obsolete"),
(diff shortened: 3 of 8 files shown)


edi_oca/models/edi_exchange_record.py
```diff
@@ -573,3 +574,19 @@ def with_delay(self, **kw):
         params = self._job_delay_params()
         params.update(kw)
         return super().with_delay(**params)
+
+    def get_fresher_duplicates(self):
```
```suggestion
    def get_fresher_duplicates(self, count=None):
```

edi_oca/models/edi_exchange_record.py
```diff
@@ -573,3 +574,19 @@ def with_delay(self, **kw):
...
+        self.ensure_one()
+        fresher_duplicates = self.search(
+            [
+                ("id", ">", self.id),
+                ("res_id", "=", self.res_id),
+                ("model", "=", self.model),
+                ("type_id", "=", self.type_id.id),
+                ("exchange_file", "!=", False),
+            ]
+        )
```
```suggestion
            ],
            count=count
        )
```

edi_oca/models/edi_exchange_record.py
```diff
@@ -573,3 +574,19 @@ def with_delay(self, **kw):
...
+        return fresher_duplicates
+
+    def has_fresher_duplicates(self):
+        return len(self.get_fresher_duplicates()) > 0
```
```suggestion
        return bool(self.get_fresher_duplicates(count=True))
```

edi_oca/models/edi_backend.py
```diff
@@ -324,10 +324,20 @@ def _output_check_send(self, exchange_record):
             raise exceptions.UserError(
                 _("Record ID=%d has no file to send!") % exchange_record.id
             )
-        return exchange_record.edi_exchange_state in [
-            "output_pending",
-            "output_error_on_send",
-        ]
+        is_superseded = (
```
simahawk: I'd move this to a computed field on the exchange record `is_obsolete`.
Also I'd move this check to a new method here `_check_obsolete and call it on `exchange_send` after the check

edi_oca/data/cron.xml
```diff
@@ -33,4 +33,20 @@
         <field name="state">code</field>
         <field name="code">model.search([])._cron_check_input_exchange_sync()</field>
     </record>
+    <record
+        id="cron_edi_backend_delete_superseded_records"
```
simahawk:
```suggestion
        id="cron_edi_backend_delete_obsolete_records"
```
I think "obsolete" makes more sense than superseded. Can you apply the same find/replace everywhere?

edi_oca/models/edi_backend.py
```diff
@@ -637,3 +647,35 @@ def _is_valid_edi_action(self, action, raise_if_not=False):
...
+
+    def _cron_delete_superseded_records(self, **kw):
+        for backend in self:
+            backend._delete_superseded_records(**kw)
+
+    def _delete_superseded_records(self, record_ids=None, **kw):
+        """
+        Go through superseded records, and delete those for which their type
+        has delete_superseded_records = True.
+        """
```
```suggestion
        """Cleanup obsolete records.
        
        Go through types with `delete_obsolete_records` flag on
        and delete their obsolete records if any.
        """
```

simahawk (changes requested):
Can you please port fix/imp from 14.0 in another PR and rebase here before moving on?
I bet we'll have many conflicts later on.
It will also improve the sending by chaining jobs.

QuocDuong1306:
I have rebased on #63

edi_oca/data/cron.xml
QuocDuong1306: Thank @simahawk , I updated, could you have a look

simahawk:
@QuocDuong1306 can you rebase?

QuocDuong1306:
Hello @simahawk , I have rebased

QuocDuong1306:
Hello, I updated this commit: https://github.com/OCA/edi-framework/pull/61/commits/0d4b9bdd1b180b06e0eab7b61aeee4d7bc3dc40e

Reason:
- Adding `exchange_file` in the domain is slow when files are stored in object storage
- we also remove useless domain `("type_id.direction", "=", "output")`, it is not actually needed, as the only criteria that matters is the `is_obsolete` flag

edi_oca/models/edi_backend.py
```diff
@@ -284,6 +284,10 @@ def exchange_send(self, exchange_record):
         check = self._output_check_send(exchange_record)
         if not check:
             return "Nothing to do. Likely already sent."
+        # In case obsolete: skip sending
```
simahawk, line 287: I would do:
* remove this check completely
* modify the msg at line 286 as "Nothing to do. Likely already sent or obsolete." to leave the door opened w/ less impact
* as for the check see below the comment on `_check_obsolete`

edi_oca/models/edi_backend.py
```diff
@@ -670,3 +674,41 @@ def _is_valid_edi_action(self, action, raise_if_not=False):
             if raise_if_not:
                 raise
             return False
+
+    def _check_obsolete(self, exchange_record):
```
simahawk, line 678: A check should never write on the record IMO.
As for this check, it won't be needed if the state changes to `obsolete` before.
I propose a different solution on the exc record. See below.

edi_oca/models/edi_backend.py
```diff
@@ -670,3 +674,41 @@ def _is_valid_edi_action(self, action, raise_if_not=False):
...
+        and delete their obsolete records if any.
+        """
+        obsolete_records = self.exchange_record_model.search(
+            self._obsolete_records_domain(record_ids=record_ids)
+        )
+        _logger.info(
+            "EDI Exchange delete records: found %d obsolete records to delete.",
+            len(obsolete_records),
+        )
+        if obsolete_records:
```
simahawk, line 702: this if is not needed

edi_oca/models/edi_exchange_record.py
```diff
@@ -603,3 +612,28 @@ def delayable(self, **kw):
 
     def _job_retry_params(self):
         return {}
+
+    def _compute_is_obsolete(self):
```
simahawk, line 616: What if we do this check immediately at creation?
Right after creation we can look for older duplicates rather than new and set their state as `obsolete` after commit.

edi_oca/models/edi_exchange_record.py
```diff
@@ -603,3 +612,28 @@ def delayable(self, **kw):
...
+
+    def get_fresher_duplicates(self, count=None):
+        self.ensure_one()
+        fresher_duplicates = self.search(
+            [
+                ("id", ">", self.id),
+                ("res_id", "=", self.res_id),
+                ("model", "=", self.model),
+                ("type_id", "=", self.type_id.id),
+                ("exchange_file", "!=", False),
```
simahawk, line 632: I'm not sure about this check on the exchange_file. 1st it might be slow and 2nd... if the file is not there yet it means that it will be generated soon and it will be the newest record. No?

simahawk (changes requested):
I would like to not include this feature in the core as it adds more complexity and I'd like to leave the door open to other type of implementations.
For this reason, I would move this feature to a new module `edi_exchange_deduplicate_oca` and apply the changes suggested.
I'd like to hear your feedback tho.

santostelmo:
@QuocDuong1306 still a draft PR ?

QuocDuong1306:
Hello @simahawk @santostelmo , sorry for the delay, I will update it ASAP...

QuocDuong1306:
> I would move this feature to a new module edi_exchange_deduplicate_oca and apply the changes suggested

Hi @simahawk @santostelmo , I moved this one to the [new PR](https://github.com/OCA/edi-framework/pull/94), could you have a look? After that, maybe this one can be closed

PR #64, opened 2024-02-01
[16.0][FIX] edi_storage_oca: FS find_files returns relative paths instead of full paths

JordiMForgeFlow:
Ex: see SFTP implementation of FS (https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/implementations/sftp.html#SFTPFileSystem)

In the method '_decode_stat' which is used to generate the info returned by 'ls', the result is build using the relative path passed to the method 'ls': `out["name"] = "/".join([parent_path.rstrip("/"), stat.filename])`

### DIFF
--- a/edi_storage_oca/models/edi_backend.py
+++ b/edi_storage_oca/models/edi_backend.py
@@ -161,9 +161,10 @@ class EDIBackend(models.Model):
         if exchange_type.exchange_file_ext:
             bits.append(r"\." + exchange_type.exchange_file_ext)
         pattern = "".join(bits)
-        full_paths = utils.find_files(self.storage_id, pattern, full_input_dir_pending)
-        pending_path_len = len(full_input_dir_pending)
-        return [p[pending_path_len:].strip("/") for p in full_paths]
+        relative_paths = utils.find_files(
+            self.storage_id, pattern, full_input_dir_pending
+        )
+        return [p.strip("/") for p in relative_paths]
 
     def _storage_new_exchange_record_vals(self, file_name):
         return {"exchange_filename": file_name, "edi_exchange_state": "input_pending"}


simahawk:
@sebalix can you review this pls?

simahawk:
Probably fixed by https://github.com/OCA/edi-framework/pull/65 ?

sebalix:
This PR doesn't address the same thing, I do not know enough about this EDI module but what I did in my PR is the replacement of `fs_storage` deprecated methods.
Now I cannot say if `find_files` should return relative or full paths... :man_shrugging:

sebalix:
@JordiMForgeFlow maybe we should include your fix in my PR (or on top of it)? Basically I moved the deprecated functions of `fs_storage` inside `edi_storage_oca` (I do not know the story, why such methods are deprecated, but we'll probably copy such convenient methods here and there, cc @simahawk )

JordiMForgeFlow:
@sebalix works for me. Should I wait until your PR gets merged or do you want to apply the commit to your PR?

simahawk:
@sebalix ping

simahawk:
BTW I've just got another weird issue on list files https://github.com/OCA/edi-framework/pull/76

nilshamerlinck:
Hello @JordiMForgeFlow 
-  `_storage_get_input_filenames` is supposed to return file names, not paths
- this change would make it return paths
- so I'm a bit curious about your case: do your `_storage_create_record_if_missing` jobs currently have file paths as arguments?
- thanks!

JordiMForgeFlow:
@nilshamerlinck with my changes I keep retuning file names. What I have noticed is that the method _find_files_ is returning relative paths instead of full paths, so without my changes when doing `p[pending_path_len:]` the file name is getting messed up.

In my first comment in this PR I am showing where I saw this happening for SFTP.

nilshamerlinck:
HI @JordiMForgeFlow 
- I'm fixing a bug in the same method and I want to make sure I'm not going to introduce a regression for others/you :)
- you say this change will keep returning file names but let's take this example:
  - backend with `input_dir_pending = "IMPORT"`
  - `find_files(".*", "IMPORT")` would return `['IMPORT/file.name']`
  - currently the code would remove the 'IMPORT/' part, but with your change it would not anymore
- that's why I'm curious about your case, if possible would you mind sharing some elements of your config:
  - storage's `directory_path`
  - backend's `input_dir_pending`
  - SFTP server name/version
- thanks!

JordiMForgeFlow:
@nilshamerlinck I have the following config:

storage's directory_path: empty
backend's input_dir_pending: input
SFTP server name/version: Version 3

In my case, the method find_files is returning `/file.name` instead of `input/file.name`

nilshamerlinck:
thanks @JordiMForgeFlow I will take it into account
sorry, a couple more questions:
- do you have an `exchange_filename_pattern` defined on your exchange type?
- do you know your SFTP server software name? (unfortunately, there seem to be some differences between implementations)

JordiMForgeFlow:
@nilshamerlinck no problem, thank you for looking into it :)

- Yes, I have an exchange_filename_pattern (example: .*INV.*)
- I don't have this information currently, the server is maintained by an external party, I can try to find out.

JordiMForgeFlow:
HI @nilshamerlinck

Have you been able to check this? From my end this is still needed, so I would hope it gets merged at some point.

JordiMForgeFlow:
@etobella having included this fix in V17 I think we can merge this one :)

etobella (approved):
/ocabot merge patch

PR #65, opened 2024-02-01
[16.0] edi_storage_oca: replace use of 'fs_storage' deprecated functions

sebalix:
Tested with SFTP protocol.

cc @simahawk

### DIFF
--- a/edi_storage_oca/components/listener.py
+++ b/edi_storage_oca/components/listener.py
@@ -7,6 +7,8 @@ from pathlib import PurePath
 
 from odoo.addons.component.core import Component
 
+from .. import utils
+
 
 class EdiStorageListener(Component):
     _name = "edi.storage.component.listener"
@@ -17,21 +19,24 @@ class EdiStorageListener(Component):
         to_dir = PurePath(to_dir_str)
         # - storage.list_files now includes path in fs_storage, breaking change
         # - we remove path
-        # TODO: clean this up, .list_files is deprecated in fs_storage
-        files = storage.list_files(from_dir.as_posix())
+        files = utils.list_files(storage, from_dir.as_posix())
         files = [os.path.basename(f) for f in files]
         if filename not in files:
             return False
-        # TODO: clean this up, .move_files is deprecated in fs_storage
         self._add_post_commit_hook(
-            storage.move_files, [(from_dir / filename).as_posix()], to_dir.as_posix()
+            utils.move_files,
+            storage,
+            [(from_dir / filename).as_posix()],
+            to_dir.as_posix(),
         )
         return True
 
-    def _add_post_commit_hook(self, move_func, sftp_filepath, sftp_destination_path):
+    def _add_post_commit_hook(
+        self, move_func, storage, sftp_filepath, sftp_destination_path
+    ):
         """Add hook after commit to move the file when transaction is over."""
         self.env.cr.postcommit.add(
-            functools.partial(move_func, sftp_filepath, sftp_destination_path)
+            functools.partial(move_func, storage, sftp_filepath, sftp_destination_path)
         )
 
     def on_edi_exchange_done(self, record):
--- a/edi_storage_oca/components/receive.py
+++ b/edi_storage_oca/components/receive.py
@@ -14,7 +14,4 @@ class EDIStorageReceiveComponent(Component):
     _usage = "storage.receive"
 
     def receive(self):
-        path = self._get_remote_file_path("pending")
-        # TODO: clean this up, .get is deprecated in fs_storage
-        filedata = self.storage.get(path.as_posix())
-        return filedata
+        return self._get_remote_file("pending", binary=True)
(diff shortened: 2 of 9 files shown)


simahawk (commented):
what about `find_files`? See #64

sebalix:
I'm suspecting (but I'm not sure) it was the use of these deprecated methods that triggers connection issues over time. These methods are decorated, and it could be that the decorator is keeping an old reference to `self.__fs` having a socket that gets closed after some time. That could explain why the "Test connection" button of `fs_storage` provides the expected result, while the EDI jobs are failing even when requeued.

This will be tested anyway.

edi_storage_oca/utils.py
```diff
@@ -0,0 +1,61 @@
...
+import base64
+import os
+import re
+
+
+def add_file(storage, path, filedata, binary=False):
+    data = filedata
+    if not binary:
+        data = base64.b64decode(filedata)
+    relative_path = path.as_posix()
```
JordiMForgeFlow: we should use `PurePath(path).as_posix()`

edi_storage_oca/utils.py
```diff
@@ -0,0 +1,61 @@
...
+    regex = re.compile(pattern)
+    for file_path in fs.ls(relative_path, detail=False):
+        # fs.ls returns a relative path
+        if regex.match(os.path.basename(file_path)):
+            result.append(file_path)
+    return result
+
+
+def get_file(storage, path, binary=False):
+    relative_path = path.as_posix()
```
JordiMForgeFlow: same issue as in the _add_file_ method

edi_storage_oca/components/receive.py
```diff
@@ -14,7 +14,4 @@ class EDIStorageReceiveComponent(Component):
     _usage = "storage.receive"
 
     def receive(self):
-        path = self._get_remote_file_path("pending")
-        # TODO: clean this up, .get is deprecated in fs_storage
-        filedata = self.storage.get(path.as_posix())
-        return filedata
+        return self._get_remote_file(self, "pending")
```
JordiMForgeFlow:
```suggestion
        return self._get_remote_file("pending")
```

edi_storage_oca/components/listener.py
```diff
@@ -17,14 +19,15 @@ def _move_file(self, storage, from_dir_str, to_dir_str, filename):
...
-        files = storage.list_files(from_dir.as_posix())
+        files = utils.list_files(storage, from_dir.as_posix())
         files = [os.path.basename(f) for f in files]
         if filename not in files:
             return False
-        # TODO: clean this up, .move_files is deprecated in fs_storage
         self._add_post_commit_hook(
-            storage.move_files, [(from_dir / filename).as_posix()], to_dir.as_posix()
+            utils.move_files,
+            self.storage,
```
JordiMForgeFlow:
```suggestion
            storage,
```

JordiMForgeFlow: _add_post_commit_hook should also be modified to receive the storage parameter

edi_storage_oca/utils.py
sebalix: I was working on this as this triggered this error during our tests: `AttributeError: 'str' object has no attribute 'as_posix'`.
I removed all calls to `as_posix` in these helpers so they now receive only paths as strings. Should they accept only `pathlib.Path` objects instead? :man_shrugging:

edi_storage_oca/components/receive.py
```diff
@@ -14,7 +14,4 @@ class EDIStorageReceiveComponent(Component):
     _usage = "storage.receive"
 
     def receive(self):
-        path = self._get_remote_file_path("pending")
-        # TODO: clean this up, .get is deprecated in fs_storage
-        filedata = self.storage.get(path.as_posix())
-        return filedata
+        return self._get_remote_file("pending")
```
JordiMForgeFlow:
```suggestion
        return self._get_remote_file("pending", binary=True)
```

JordiMForgeFlow (approved):
Code + functional review 👍🏼

edi_storage_oca/utils.py
```diff
@@ -0,0 +1,60 @@
+# Copyright 2024 Camptocamp SA
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl)
+
+import base64
+import os
+import re
+
+
+def add_file(storage, path, filedata, binary=False):
```
simahawk, line 9: nice. This is something that should probably be proposed to the base module as I don't see any edi specific stuff.

simahawk (approved):

simahawk:
@sebalix can you squash pls?

edi_storage_oca/utils.py
sebalix: It was in the base module `fs_storage` before (as Odoo methods, not functions), but deprecated. Thing is we still need to encapsulate such features somewhere yes.

PR #73, opened 2024-02-28
[16.0][IMP] edi_storage_oca: new opt-in technical field to avoid moving files after processing

thienvh332:
In some contexts, we might have an EDI Exchange Type with both:

- a storage (e.g. files pushed to a SFTP)

- AND an endpoint

If a record is created through the endpoint, we want to disable the logic that moves files after processing (implemented as a listener:
https://github.com/oca/edi-framework/blob/16.0/edi_storage_oca/components/listener.py#L13)

### DIFF
--- a/edi_storage_oca/models/edi_backend.py
+++ b/edi_storage_oca/models/edi_backend.py
@@ -139,7 +139,8 @@ class EDIBackend(models.Model):
         if existing:
             return
         record = self.create_record(
-            exchange_type.code, self._storage_new_exchange_record_vals(file_name)
+            exchange_type.code,
+            self._storage_new_exchange_record_vals(file_name),
         )
         _logger.debug("%s: new exchange record generated.", self.name)
         return record.identifier
@@ -166,4 +167,8 @@ class EDIBackend(models.Model):
         return [p[pending_path_len:].strip("/") for p in full_paths]
 
     def _storage_new_exchange_record_vals(self, file_name):
-        return {"exchange_filename": file_name, "edi_exchange_state": "input_pending"}
+        return {
+            "exchange_filename": file_name,
+            "edi_exchange_state": "input_pending",
+            "storage_id": self.storage_id.id,
+        }
--- a/edi_storage_oca/models/edi_exchange_record.py
+++ b/edi_storage_oca/models/edi_exchange_record.py
@@ -1,0 +1,15 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class EDIExchangeRecord(models.Model):
+    _inherit = "edi.exchange.record"
+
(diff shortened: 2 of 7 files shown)


edi_storage_oca/models/edi_exchange_record.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class EDIExchangeRecord(models.Model):
+    _inherit = "edi.exchange.record"
+
+    move_file_after_processing = fields.Boolean(
```
simahawk: In the endpoint module there's a reference to the endpoint that originated the record.
I would rather keep a reference to the storage that originated the record. 
In any case, the value must be filled automatically by the storage cron.

simahawk (changes requested):

edi_storage_oca/models/edi_exchange_record.py
thienvh332: Thanks @simahawk , I updated it

edi_storage_oca/models/edi_backend.py
```diff
@@ -165,5 +167,9 @@ def _storage_get_input_filenames(self, exchange_type):
         pending_path_len = len(full_input_dir_pending)
         return [p[pending_path_len:].strip("/") for p in full_paths]
 
-    def _storage_new_exchange_record_vals(self, file_name):
-        return {"exchange_filename": file_name, "edi_exchange_state": "input_pending"}
+    def _storage_new_exchange_record_vals(self, file_name, storage_id):
```
simahawk: Rule of thumb: always try to not change the signature of an existing method w/ a required param.
In my opinion here is not useful anyway because this is an instance method and you can still use self.storage_id inside it. Do you see any problem w/ that?

simahawk (changes requested):

edi_storage_oca/models/edi_backend.py
thienvh332: I updated it. Thanks @simahawk

edi_storage_oca/models/edi_exchange_record.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2024 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class EDIExchangeRecord(models.Model):
+    _inherit = "edi.exchange.record"
+
+    storage_id = fields.Many2one(
```
simahawk, line 10: Sorry, I missed on step: for consistency this field should be automatically populated for existing records via a proper sql query to update them all at once on post-migrate using their `backend_id.storage_id`.

simahawk (changes requested):

simahawk (approved):

PR #74, opened 2024-03-06
[IMP] edi_oca: add a backend_type_code field on the backend

gurneyalex:
this field can be used in backend views to show/hide configuration elements depending on the backend type


simahawk (approved):

etobella (approved):
/ocabot merge patch

PR #76, opened 2024-03-07
[16.0] edi_storage: fix list_files

simahawk:
TODO: explanation... don't know yet why this is happening for certain SFTP server only.

With `detail=False` it breaks like this:
```
>>> fs.ls(fpath.as_posix(), detail=False)
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<console>", line 1, in <module>
  File "/usr/local/lib/python3.10/site-packages/fsspec/implementations/dirfs.py", line 255, in ls
    return self._relpath(ret)
  File "/usr/local/lib/python3.10/site-packages/fsspec/implementations/dirfs.py", line 70, in _relpath
    return [self._relpath(_path) for _path in path]
  File "/usr/local/lib/python3.10/site-packages/fsspec/implementations/dirfs.py", line 70, in <listcomp>
    return [self._relpath(_path) for _path in path]
  File "/usr/local/lib/python3.10/site-packages/fsspec/implementations/dirfs.py", line 68, in _relpath
    assert path.startswith(prefix)
AssertionError
```

### DIFF
--- a/edi_storage_oca/utils.py
+++ b/edi_storage_oca/utils.py
@@ -26,7 +26,8 @@ def find_files(storage, pattern, relative_path="", **kw) -> list[str]:
     if not fs.exists(relative_path):
         return []
     regex = re.compile(pattern)
-    for file_path in fs.ls(relative_path):
+    for item in fs.ls(relative_path):
+        file_path = item["name"]
         # fs.ls returns a relative path
         if regex.match(os.path.basename(file_path)):
             result.append(file_path)
--- a/edi_storage_oca/utils.py
+++ b/edi_storage_oca/utils.py
@@ -48,7 +48,7 @@ def list_files(storage, relative_path="", pattern=False):
     if pattern:
         relative_path = fs.sep.join([relative_path, pattern])
         return fs.glob(relative_path)
-    return fs.ls(relative_path, detail=False)
+    return fs.ls(relative_path)
 
 
 def move_files(storage, files, destination_path, **kw):
--- a/edi_storage_oca/utils.py
+++ b/edi_storage_oca/utils.py
@@ -26,7 +26,7 @@ def find_files(storage, pattern, relative_path="", **kw) -> list[str]:
     if not fs.exists(relative_path):
         return []
     regex = re.compile(pattern)
-    for file_path in fs.ls(relative_path, detail=False):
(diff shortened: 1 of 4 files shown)


edi_storage_oca/utils.py
```diff
@@ -48,7 +49,7 @@ def list_files(storage, relative_path="", pattern=False):
     if pattern:
         relative_path = fs.sep.join([relative_path, pattern])
         return fs.glob(relative_path)
```
nilshamerlinck, line 51:
```suggestion
        return fs.glob(relative_path, detail=True)
```

edi_storage_oca/utils.py
```diff
@@ -48,7 +49,7 @@ def list_files(storage, relative_path="", pattern=False):
...
-    return fs.ls(relative_path, detail=False)
+    return fs.ls(relative_path)
```
nilshamerlinck, line 52:
```suggestion
    return fs.ls(relative_path, detail=True)
```

edi_storage_oca/utils.py
```diff
@@ -26,7 +26,8 @@ def find_files(storage, pattern, relative_path="", **kw) -> list[str]:
     if not fs.exists(relative_path):
         return []
     regex = re.compile(pattern)
-    for file_path in fs.ls(relative_path, detail=False):
+    for item in fs.ls(relative_path):
```
nilshamerlinck, line 29:
```suggestion
    for item in fs.ls(relative_path, detail=True):
```

nilshamerlinck (changes requested):
hi, we need an explicit `detail=True`
- not only to make clear that expected output here is a list of dicts
- but also because some underlying implementations have `detail=False` for default, e.g. [sftp](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/sftp.py#L125), so removing `detail=False` is not enough to get expected `detail=True`

simahawk:
@nilshamerlinck unfortunately this is broken w/ the SFTP used. We can check together if you want.

simahawk:
@nilshamerlinck feel free to take over pls :pray:

simahawk:
@nilshamerlinck ping :)

PR #77, opened 2024-03-26
[16.0][IMP] edi_storage_oca: Fix permission

tuantrantg:
We want internal users to be able to trigger creation of EDI records without being part of EDI Manager group and without having read right on `fs.storage`


simahawk:
LG, can you please clean the commit msg?

```
edi_storage_oca: fix permission

...bla bla... explanation...
```

tuantrantg:
> LG, can you please clean the commit msg?
> 
> ```
> edi_storage_oca: fix permission
> 
> ...bla bla... explanation...
> ```

hi @simahawk please review it again

simahawk (approved):

PR #79, opened 2024-04-01
[IMP] edi_oca: Improve archive/unarchive exchange type rules

QuocDuong1306:
Change of behavior:
  - When an exchange type is archived, its exchange type rules are also archived (same as before)
  - When an exchange type is re-activated, its exchange type rules are also re-activated (new)

### DIFF
--- a/edi_oca/models/edi_exchange_type.py
+++ b/edi_oca/models/edi_exchange_type.py
@@ -30,7 +30,7 @@ class EDIExchangeType(models.Model):
     _name = "edi.exchange.type"
     _description = "EDI Exchange Type"
 
-    active = fields.Boolean(default=True, inverse="_inverse_active")
+    active = fields.Boolean(default=True)
     backend_id = fields.Many2one(
         string="Backend",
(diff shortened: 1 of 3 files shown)


etobella:
I don't like both of the approaches... If we do it this way, if I inactivate a rule and then I inactivate and activate the type, the rule will be active again... why don't we keep it separated and improve the search of rules by checking that the rule and the type is active?

santostelmo (approved):

simahawk (changes requested):
> I don't like both of the approaches... If we do it this way, if I inactivate a rule and then I inactivate and activate the type, the rule will be active again... 

I agree. I did it on purpose for this reason.

> why don't we keep it separated and improve the search of rules by checking that the rule and the type is active?

The rules are already searched and the active flag should play is role (unless you enforce active_test) in the ctx -> we could enforce this in the domain maybe.

QuocDuong1306:
Thanks for your suggestion, I think this improvement is not needed, so I close

PR #84, opened 2024-05-13
[16.0][IMP] edi_oca: allow to receive empty files

QuocDuong1306:
Currently when we receive an empty file:
- Receive step doesn't end up in state error_on_receive : that's because we just ignore the content [here](https://github.com/OCA/edi-framework/blob/77184146551ea042190ca7b0a1c290f848dd202f/edi_oca/models/edi_backend.py#L530)
- But the [process step will raise an error](https://github.com/OCA/edi-framework/blob/77184146551ea042190ca7b0a1c290f848dd202f/edi_oca/models/edi_backend.py#L460)
- So this PR goal tries to improve that
- By default, this option is disabled so we still consider that an empty file is an error case

### DIFF
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -272,6 +272,12 @@ class EDIBackend(models.Model):
 
     # TODO: add tests
     def _validate_data(self, exchange_record, value=None, **kw):
+        if exchange_record.direction == "input" and not exchange_record.exchange_file:
+            if not exchange_record.type_id.allow_empty_files_on_receive:
+                raise ValueError(
+                    _("Empty files are not allowed for this exchange type")
+                )
+
         component = self._get_component(exchange_record, "validate")
         if component:
             return component.validate(value)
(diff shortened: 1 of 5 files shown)


edi_oca/models/edi_backend.py
```diff
@@ -579,6 +580,21 @@ def _exchange_receive(self, exchange_record):
             return component.receive()
         raise NotImplementedError()
 
+    def _process_received_content(self, exchange_record, content):
```
simahawk: IMO this should be done in `validate_data` and you can set the content of the file directly.
I would:
* keep the code as it was before but remove the check on the content value.
* set the value as it is on the record
* call validate data and raise the error there if necessary

edi_oca/models/edi_backend.py
```diff
@@ -579,6 +580,21 @@ def _exchange_receive(self, exchange_record):
...
+        # Do nothing when the content is None (result of FileNotFoundError/OSError)
+        if content is None:
+            return None
+
+        if len(content) == 0:
+            if not exchange_record.type_id.allow_empty_files_on_receive:
+                raise ValueError(
+                    "File is Empty and Empty Files are not allowed for this exchange type"
```
```suggestion
                    _("Empty files are not allowed for this exchange type")
```

simahawk (changes requested):

simahawk:
`(disabled by default)` should be moved to the commit description together w/ the PR description :wink:

QuocDuong1306:
Hi @simahawk , I updated the PR

simahawk (approved):

etobella (approved):

simahawk:
Ooops... I hit the merge btn by error :sweat:

PR #70, opened 2024-02-24
[16.0][IMP] edi_storage_oca: do not ignore OSError when getting remote file

JordiMForgeFlow:
When the fs_storage connection fails, it can raise OSError here: https://github.com/OCA/storage/blob/4d82d00aaec818c10dfcfe509d9640d48ebffbf2/fs_storage/models/fs_storage.py#L294.

PR #88, opened 2024-06-12
[16.0][FIX] edi_storage_oca: do not silently ignore OSError/FileNotFoundError

nilshamerlinck:
Depends on https://github.com/OCA/queue/pull/674

- if an `OSError` is raised, there can be different cases but it's likely a transient issue (e.g. connection error):
  - if we ignore it as currently:
    - Job won't fail
    - the exchange record will be incorrectly marked as "input_received" but with no data, which is not what we want
  - so we don't ignore it anymore:
    - Job will fail;
    - ~~the record state will stay as "input_pending" (as exception won't be catched in [edi.backend:exchange_receive](https://github.com/OCA/edi-framework/blob/16.0/edi_oca/models/edi_backend.py#L537))~~
    - ~~and as such, another `action_exchange_receive` Job will be created next time scheduled action "EDI exchange check input sync" runs, to try again to receive it, hopefully successfully this time~~
    - a `RetryableJobError` will be raised in order to retry it
    - if it sill can't succeed after `max_retries`, record state will be changed to "input_receive_error"

- if a `FileNotFoundError` is raised, it's not a transient issue, so record should be marked as "input_receive_error":
  - to do so we don't ignore it anymore:
    - it will get catched in [edi.backend:exchange_receive](https://github.com/OCA/edi-framework/blob/16.0/edi_oca/models/edi_backend.py#L537) as a ["swallable"](https://github.com/OCA/edi-framework/blob/16.0/edi_oca/models/edi_backend.py#L548) (Final) error
    - record state will be changed to "input_receive_error"

### DIFF
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -539,7 +539,6 @@ class EDIBackend(models.Model):
         content = None
         try:
             content = self._exchange_receive(exchange_record)
-            # Ignore result of FileNotFoundError/OSError
             if content is not None:
                 exchange_record._set_file_content(content)
                 self._validate_data(exchange_record)
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -348,11 +348,20 @@ class EDIBackend(models.Model):
             exceptions.ValidationError,
         )
 
-    def _send_retryable_exceptions(self):
+    def _common_retryable_exceptions(self):
         # IOError is a base class for all connection errors
         # OSError is a base class for all errors
         # when dealing w/ internal or external systems or filesystems
         return (IOError, OSError)
+
+    def _receive_retryable_exceptions(self):
+        return self._common_retryable_exceptions()
+
+    def _process_retryable_exceptions(self):
+        return self._common_retryable_exceptions()
+
+    def _send_retryable_exceptions(self):
+        return self._common_retryable_exceptions()
 
     def _output_check_send(self, exchange_record):
         if exchange_record.direction != "output":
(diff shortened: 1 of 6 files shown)


JordiMForgeFlow:
@nilshamerlinck is there a reason why you don't keep the logging? Besides that, the changes look good to me and solve https://github.com/OCA/edi-framework/pull/70

nilshamerlinck:
good catch @JordiMForgeFlow I can definitely keep it for the `raise_if_not_found=False` case, fixed!

JordiMForgeFlow (approved):

santostelmo (approved):

simahawk (changes requested):
LG. Could you please take the chance to add a test for this? :pray: 
Also, one consideration: 
> another action_exchange_receive Job will be created next time scheduled action "EDI exchange check input sync" runs, to try again to receive it, hopefully successfully this time

Wouldn't be better to raise a `RetryableJobError` then, instead of getting another job?

Of course, if we go this way we should go ahead w/ the fix on queue_job we discussed together regarding matching existing jobs when on error+retry counter>0 :wink:

simahawk:
@nilshamerlinck ping :)

nilshamerlinck:
Hello @simahawk 
- @QuocDuong1306 just pushed commits that will apply the same retry logic for _receive+process_ as for _send_: rely on the _RetryableJobError_ mechanism instead of the scheduled action ("EDI exchange check input sync"), as per your suggestion
- regarding the fix on queue_job, we went for a different approach:
  - instead of preventing the creation of a new job if a failed job with `max_retries` already exists
  - we explicitly mark the exchange record as KO through a hook, see https://github.com/OCA/queue/pull/674
- thoughts welcome!

santostelmo:
Moved this PR to https://github.com/OCA/edi-framework/pull/150

PR #89, opened 2024-06-12
[16.0][FIX] edi_oca: do not try to automatically re-process input_processed_error records

nilshamerlinck:
- scheduled action "EDI exchange check input sync" will currently create `action_exchange_process` jobs for any record in state "input_processed_error"
- this does not make any sense, as such records failed to process because of ["swallable"](https://github.com/OCA/edi-framework/blob/16.0/edi_oca/models/edi_backend.py#L490) exceptions, that should be considered as final

### DIFF
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -633,7 +633,7 @@ class EDIBackend(models.Model):
         return domain
 
     def _input_pending_process_records_domain(self, record_ids=None):
-        states = ("input_received", "input_processed_error")
+        states = ("input_received",)
         domain = [
             ("backend_id", "=", self.id),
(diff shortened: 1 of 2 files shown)


simahawk (changes requested):
makes sense, but I'd like to see a test :wink:

QuocDuong1306:
Hello @simahawk @nilshamerlinck , I added a test and rebased also

simahawk (approved):
LGTM. @etobella WDYT?

We should back/fwd port this when ready.

etobella (approved):

simahawk:
/ocabot merge patch

@nilshamerlinck would you mind fwd/back porting? :pray:

PR #93, opened 2024-06-28
[16.0][IMP] edi_oca: add smart button in record form to show all related queue jobs

QuocDuong1306:
This PR adds a smart button in record form to show all related queue jobs

Depend on:
- https://github.com/OCA/edi/pull/1004

### DIFF
--- a/edi_oca/models/edi_exchange_record.py
+++ b/edi_oca/models/edi_exchange_record.py
@@ -5,6 +5,7 @@
 
 import base64
 import logging
+from ast import literal_eval
 from collections import defaultdict
 
 from odoo import _, api, exceptions, fields, models
@@ -112,6 +113,9 @@ class EDIExchangeRecord(models.Model):
     retryable = fields.Boolean(
         compute="_compute_retryable",
         help="The record state can be rolled back manually in case of failure.",
+    )
+    related_queue_jobs_count = fields.Integer(
+        compute="_compute_related_queue_jobs_count"
     )
     company_id = fields.Many2one("res.company", string="Company")
 
@@ -610,3 +614,32 @@ class EDIExchangeRecord(models.Model):
 
     def _job_retry_params(self):
         return {}
+
+    def _compute_related_queue_jobs_count(self):
+        for rec in self:
+            # TODO: We should refactor the object field on queue_job to use jsonb field
+            # so that we can search directly into it.
+            rec.related_queue_jobs_count = rec.env["queue.job"].search_count(
+                [("func_string", "like", str(rec))]
+            )
+
+    def action_view_related_queue_jobs(self):
+        self.ensure_one()
+        xmlid = "queue_job.action_queue_job"
+        action = self.env["ir.actions.act_window"]._for_xml_id(xmlid)
+        # Searching based on task name. Ex: `edi.exchange.record(1,).action_exchange_send()`
+        # TODO: We should refactor the object field on queue_job to use jsonb field
+        # so that we can search directly into it.
+        action["domain"] = [("func_string", "like", str(self))]
+        # Purge default search filters from ctx to avoid hiding records
+        ctx = action.get("context", {})
+        if isinstance(ctx, str):
+            ctx = literal_eval(ctx)
+        # Update the current contexts
+        ctx.update(self.env.context)
+        action["context"] = {
+            k: v for k, v in ctx.items() if not k.startswith("search_default_")
+        }
+        # Drop ID otherwise the context will be loaded from the action's record
+        action.pop("id")
+        return action
--- a/edi_oca/readme/CONFIGURE.rst
+++ b/edi_oca/readme/CONFIGURE.rst
@@ -10,6 +10,15 @@ In order to define a new Exchange Record, we need to configure:
 * Backend
 * Components
 
+Jobs
+~~~~~~~~~~~~~~~~~~~~
+
+* (1) **Internal User**: might be an EDI user without even knowing about it, triggering EDI flows by some of his actions on business records; does not need access to related queue jobs.
+
+* (2) **EDI User**: more conscious EDI user that might sometimes need to debug things a bit further and thus needs access to related queue jobs.
+
+* (3) **EDI Manager**: full configuration access.
+
 Component definition
 ~~~~~~~~~~~~~~~~~~~~
 
--- a/edi_oca/readme/CONFIGURE.rst
+++ b/edi_oca/readme/CONFIGURE.rst
@@ -11,7 +11,7 @@ In order to define a new Exchange Record, we need to configure:
 * Components
 
 Jobs
-~~~~~~~~~~~~~~~~~~~~
+~~~~
 
 * (1) **Internal User**: might be an EDI user without even knowing about it, triggering EDI flows by some of his actions on business records; does not need access to related queue jobs.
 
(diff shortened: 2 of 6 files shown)


edi_oca/models/edi_exchange_record.py
```diff
@@ -603,3 +607,28 @@ def delayable(self, **kw):
 
     def _job_retry_params(self):
         return {}
+
+    def _compute_related_queue_jobs_count(self):
+        for rec in self:
+            rec.related_queue_jobs_count = rec.env["queue.job"].search_count(
+                [("func_string", "like", str(rec))]
```
simahawk, line 623: Can we add a TODO? We should refactor the object field on queue_job to use jsonb field so that we can search directy into it.

edi_oca/models/edi_exchange_record.py
```diff
@@ -603,3 +607,28 @@ def delayable(self, **kw):
...
+            )
+
+    def action_view_related_queue_jobs(self):
```
simahawk, line 626: I'd like to see a test for the action and the counter. Thanks!

edi_oca/tests/test_security.py
```diff
@@ -113,7 +113,15 @@ def test_no_group_no_read(self):
     @mute_logger("odoo.addons.base.models.ir_rule")
     def test_rule_no_read(self):
         exchange_record = self.create_record()
-        self.user.write({"groups_id": [(4, self.group.id)]})
+        self.user.write(
+            {
+                "groups_id": [
+                    # To read the related queue jobs
+                    (4, self.env.ref("queue_job.group_queue_job_manager").id),
```
simahawk: This is a sign that the avg user will probably get an error when jumping to the record, no?
Shall we introduce a basic edi user group and imply the read access on jobs for it?
As an alternative I would protect the field and the action with the manager group.

simahawk (changes requested):

edi_oca/tests/test_backend_jobs.py
```diff
@@ -31,6 +37,9 @@ def test_output(self):
         self.assertEqual(
             created.name, "Generate output content for given exchange record."
         )
+        # Check related jobs
+        self.assertEqual(len(created), record.related_queue_jobs_count)
```
simahawk: This is redundant as you compare the records below. Same for all the occurrences.

QuocDuong1306:
(Waiting for merging the dependent PR, will update if ended)

QuocDuong1306:
Hello @simahawk , I updated as you suggested, could you have a look?

edi_oca/tests/test_backend_jobs.py
QuocDuong1306: Thank @simahawk , Is it better yet?

simahawk (approved):

edi_oca/security/ir_model_access.xml
```diff
@@ -45,10 +45,22 @@
         <field name="perm_write" eval="1" />
         <field name="perm_unlink" eval="1" />
     </record>
+
+    <!-- Access right to read related jobs on record -->
+    <record model="ir.model.access" id="access_queue_job_user">
+        <field name="name">access_queue_job user</field>
```
simahawk:
```suggestion
        <field name="name">access_queue_job edi user</field>
```

QuocDuong1306: Updated, thank @simahawk

QuocDuong1306:
Hi @simahawk , I pushed 1 fixup commit for cases:

Assume that DB has already installed `edi_sale_oca`, `edi_stock_oca`, `edi_product_oca`, `edi_partner_oca`.. . On their form (sale.order, stock.picking, product.template,...) will limit the access right because their group was not implied `base_edi.group_edi_user` to read records on [edi.exchange.consumer.mixin](https://github.com/OCA/edi-framework/blob/e9a950a15891dc232cd5deb8fe8928b6248cbe19/edi_oca/models/edi_exchange_consumer_mixin.py#L20) (big problem because of breaking flows and tests)

I suggest that:
 - Keep creating access rights to EDI stack for `base.group_user`, but should create read `queue.job` permission for `base_edi.group_edi_user`, for that edi user can access to `edi.exchange.record` form (not accepted for `base.group_user`)
 
Ex: Break when accessing to `res.partner` if users have not implied edi user group
image

simahawk:
> Hi @simahawk , I pushed 1 fixup commit for cases:
> 
> Assume that DB has already installed `edi_sale_oca`, `edi_stock_oca`, `edi_product_oca`, `edi_partner_oca`.. . On their form (sale.order, stock.picking, product.template,...) will limit the access right because their group was not implied `base_edi.group_edi_user` to read records on [edi.exchange.consumer.mixin](https://github.com/OCA/edi-framework/blob/e9a950a15891dc232cd5deb8fe8928b6248cbe19/edi_oca/models/edi_exchange_consumer_mixin.py#L20) (big problem because of breaking flows and tests)
> 
> I suggest that:
> 
>     * Keep creating access rights to EDI stack for `base.group_user`, but should create read `queue.job` permission for `base_edi.group_edi_user`, for that edi user can access to `edi.exchange.record` form (not accepted for `base.group_user`)
> 
> 
> Ex: Break when accessing to `res.partner` if users have not implied edi user group image

Can't we have a migration step to fix this?
If not, let's go as you propose but we should some how log a deprecation.

nilshamerlinck:
Hi @simahawk do you agree with these levels of access:
 - (1) **Internal User**: might be an EDI user without even knowing about it, triggering EDI flows by some of his actions on business records; does not need access to related queue jobs
 - (2) (new) **EDI User**: more conscious EDI user that might sometimes need to debug things a bit further and thus needs access to related queue jobs 
 - (3) **EDI Manager**: full configuration access
 
 If yes, then this PR does not introduce any change, just implements (2)

simahawk:
> Hi @simahawk do you agree with these levels of access:
> 
>     * (1) **Internal User**: might be an EDI user without even knowing about it, triggering EDI flows by some of his actions on business records; does not need access to related queue jobs
> 
>     * (2) (new) **EDI User**: more conscious EDI user that might sometimes need to debug things a bit further and thus needs access to related queue jobs
> 
>     * (3) **EDI Manager**: full configuration access
> 
> 
> If yes, then this PR does not introduce any change, just implements (2)

@nilshamerlinck thanks for the summary! I'm ok w/ this then.
@QuocDuong1306 could you please add this info somewhere? Maybe a section "Jobs" in "CONFIGURE.rst" is appropriate.
@etobella could you please pass by for a review?

QuocDuong1306:
Hello @simahawk , the PR is updated

santostelmo (approved):

santostelmo (approved):

MiquelRForgeFlow:
conflicts

edi_oca/readme/CONFIGURE.rst
```diff
@@ -10,6 +10,15 @@ In order to define a new Exchange Record, we need to configure:
 * Backend
 * Components
 
+Jobs
+~~~~~~~~~~~~~~~~~~~~
```
simahawk:
```suggestion
~~~~
```
This won't probably pass a strict RST validation

edi_oca/static/description/index.html
```diff
@@ -8,11 +8,10 @@
```
simahawk: TIP for the future: you can use `export SKIP=oca-gen-addon-readme` in your shell to avoid this noise  and possible conflicts :wink: 
See https://github.com/OCA/edi-framework/blob/16.0/.github/workflows/pre-commit.yml#L38

simahawk: I cleaned up the commits. Will backport  again to https://github.com/OCA/edi/pull/1029 once merged

simahawk (approved):

PR #95, opened 2024-07-08
[16.0][IMP] edi_storage_oca: Add script to migrate to v16

QuocDuong1306:
### Context:
- Assume that `storage_backend` will still be installed in the old db and that we simply have to create the matching `fs.storage` records w/ their conf leaving the old storage.backend records untouched.

### Notes:
- Put this script to `16.0.1.0.0` because of changing storage_backend to fs_storage, if we put it to the later versions, the storages in backends have been changed, so can't link to the mapping records in `fs.storage`

- `storage_backend_ftp` and `storage_backend_s3` have not been migrated to 16.0 (and should not)

- `storage_backend` depends on `server_environment`
   - we just need to create a fs_storage, and link it with the edi_backend
   - the rest (host/port/credentials) will come from the environment
   
  cc @simahawk @etobella

### DIFF
--- a/edi_storage_oca/migrations/16.0.1.0.0/pre-migrate.py
+++ b/edi_storage_oca/migrations/16.0.1.0.0/pre-migrate.py
@@ -1,0 +1,79 @@
+# Copyright 2024 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+from odoo.tools.sql import create_column
+
+from odoo.addons.http_routing.models.ir_http import slugify
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if not version:
+        return
+
+    required_modules = ("storage_backend", "edi_storage_oca")
+    if not env["ir.module.module"].search(
+        [("name", "in", required_modules), ("state", "=", "installed")]
+    ):
+        return
+    if not openupgrade.column_exists(env.cr, "edi_backend", "new_storage_id"):
+        create_column(env.cr, "edi_backend", "new_storage_id", "int4")
+
+    _create_fs_storage_records(env)
+
+    if openupgrade.column_exists(env.cr, "edi_backend", "new_storage_id"):
+        _column_renames = {
+            "edi_backend": [
+                ("storage_id", "old_storage_id"),
+                ("new_storage_id", "storage_id"),
+            ],
+        }
+        openupgrade.rename_columns(env.cr, _column_renames)
+
+
+def _create_fs_storage_records(env):
+    # Create a fs.storage record for each backend.storage
+    storage_backend_records = env["storage.backend"].search([])
+    if not storage_backend_records:
+        return
+
+    # make sure all backend_type can be mapped even if corresponding modules
+    # have not been migrated (on purpose because we should switch to fs_storage)
+    selection = [
+        ("filesystem", "Filesystem"),
+        ("ftp", "FTP"),
+        ("sftp", "SFTP"),
+        ("s3", "S3"),
+    ]
+    env["storage.backend"]._fields["backend_type"].selection = selection
+    env["storage.backend"]._fields["backend_type_env_default"].selection = selection
+
+    fs_storage = env["fs.storage"]
+    for record in storage_backend_records:
+        protocol = "file"
+        if record.backend_type == "ftp":
+            protocol = "ftp"
+        elif record.backend_type == "sftp":
+            protocol = "sftp"
+        elif record.backend_type == "s3":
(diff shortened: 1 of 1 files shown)


edi_storage_oca/migrations/16.0.1.0.0/pre-migrate.py
```diff
@@ -0,0 +1,71 @@
+# Copyright 2024 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import SUPERUSER_ID, api
+from odoo.tools.sql import column_exists, create_column, rename_column
+
+from odoo.addons.http_routing.models.ir_http import slugify
+
+
+def migrate(cr, version):
```
etobella: You can use @openupgrade.migrate()
in order to have env directly on the migration function

you can check functionalities of openupgrade here:

https://github.com/OCA/OpenUpgrade/blob/16.0/openupgrade_scripts/scripts/account/16.0.1.2/pre-migration.py#L525-L528

etobella:
@luisDIXMIT

edi_storage_oca/migrations/16.0.1.0.0/pre-migrate.py
QuocDuong1306: Thank @etobella ,the PR is updated as you suggested

simahawk:
@etobella would you be able to test this in real life?

simahawk:
> * `storage_backend_ftp` and `storage_backend_s3` have not been migrated to 16.0 (and should not)

FTR I doubt we should not :wink:

etobella:
> * the rest (host/port/credentials) will come from the environment

That is not true. We could define information directly on the storage. this information needs to be ported too

edi_storage_oca/migrations/16.0.1.0.0/pre-migrate.py
```diff
@@ -0,0 +1,79 @@
...
+from odoo.addons.http_routing.models.ir_http import slugify
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if not version:
+        return
+
+    required_modules = ("storage_backend", "edi_storage_oca")
+    if not env["ir.module.module"].search(
```
etobella, line 17: I think you don't need this verificatiion. If you are in this script, you come from 15, so `storage_backend` is installed

edi_storage_oca/migrations/16.0.1.0.0/pre-migrate.py
```diff
@@ -0,0 +1,79 @@
...
+                ("storage_id", "old_storage_id"),
+                ("new_storage_id", "storage_id"),
+            ],
+        }
+        openupgrade.rename_columns(env.cr, _column_renames)
+
+
+def _create_fs_storage_records(env):
+    # Create a fs.storage record for each backend.storage
+    storage_backend_records = env["storage.backend"].search([])
```
etobella, line 38: I think we should only transform the backends that come from edi.backends. Probably we need to add some SQL here, but it is the cleaner option IMO

edi_storage_oca/migrations/16.0.1.0.0/pre-migrate.py
```diff
@@ -0,0 +1,79 @@
...
+def migrate(env, version):
+    if not version:
+        return
+
+    required_modules = ("storage_backend", "edi_storage_oca")
+    if not env["ir.module.module"].search(
+        [("name", "in", required_modules), ("state", "=", "installed")]
+    ):
+        return
+    if not openupgrade.column_exists(env.cr, "edi_backend", "new_storage_id"):
```
etobella, line 21: In order to avoid the create new column, change column to old and new to current you can do:
```python
        _column_renames = {
            "edi_backend": [
                ("storage_id", False),            ],
        }
        openupgrade.rename_columns(env.cr, _column_renames)
        create_column(env.cr, "edi_backend", "storage_id", "int4")
```
The old field will have the following name: `openupgrade.get_legacy_name("storage_id")`

nilshamerlinck:
Hi @simahawk precision about that point:
> > storage_backend_ftp and storage_backend_s3 have not been migrated to 16.0 (and should not)

> FTR I doubt we should not

of course some people might decide to migrate them, like `storage_backend_sftp` was
but what we meant here is that `fs_storage` is now the preferred way
and as concretely these modules are not available in 16.0, we needed a small trick to allow the migration script to manipulate `storage.backend` of various types, see [here](https://github.com/OCA/edi-framework/pull/95/files#diff-01e908a8f79629c0bec20d811850bbf5d71f375418e9cdef2d6e02245b9b9cf6R42-R51):

image

PR #101, opened 2024-07-29
[16.0][ADD] edi_utm_oca

chaule97:


edi_utm_oca/models/edi_exchange_type.py
```diff
@@ -0,0 +1,22 @@
...
+    _inherit = "edi.exchange.type"
+
+    edi_source_id = fields.Many2one(
+        "utm.source",
+        default=lambda self: self._default_edi_source_id(),
+        help="The UTM Source to be used by this type of exchanges.",
+    )
+
+    @api.model
+    def _default_edi_source_id(self):
```
simahawk, line 18: I don't think we need a default here.

edi_utm_oca/models/edi_utm_consumer_mixin.py
```diff
@@ -0,0 +1,27 @@
...
+from odoo import fields, models
+
+
+class EdiUtmConsumerMixin(models.AbstractModel):
+    """Provide EDI source id for related records."""
+
+    _name = "edi.utm.consumer.mixin"
+    _description = __doc__
+
+    edi_source_id = fields.Many2one("utm.source")
```
simahawk, line 14: I'm not sure to understand what's the usage of this field.
I would expect that we populate the base fields for UTM...

simahawk (changes requested):

edi_utm_oca/models/edi_utm_consumer_mixin.py
chaule97: I also added the module `edi_sale_utm_oca` and inherited this mixin [here](https://github.com/OCA/edi-framework/blob/8f08d7c2d1a4566c236283ad4f3b19933030eb78/edi_sale_utm_oca/models/sale_order.py) to save the source ID from `edi.exchange.type`. I follow up your [requirement](https://github.com/camptocamp/alloboissons_odoo/pull/758/files#r1656614436)

PR #102, opened 2024-07-29
[16.0][ADD] edi_utm_oca

chaule97:
-  this module `edi_utm_oca` automatically sets the EDI source on models that have a source_id field when they are created.


simahawk (changes requested):
As reported on the base PR I don't think we need yet another source discriminator.
We already have endpoint and storage as sources + the types, what else do we need?
Plus, we can leverage UTM for a more general classification... I fail to understand what's the whole point of this.

edi_sale_utm_oca/models/edi_exchange_type.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+from odoo import api, fields, models
+
+
+class EDIExchangeType(models.Model):
+    _inherit = "edi.exchange.type"
+
+    edi_source_id = fields.Many2one(
```
simahawk: Sorry but this is wrong. Ping @nilshamerlinck

simahawk (changes requested):
also this module should be generic and not related to sales only

edi_sale_utm_oca/models/edi_exchange_type.py
nilshamerlinck: what if someone wants a different utm.source per exchange type? or do you mean that it's the naming that is confusing and it should better be named source_id?

edi_utm_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+
+class EDIExchangeConsumerMixin(models.AbstractModel):
+    _inherit = "edi.exchange.consumer.mixin"
+
+    def _edi_update_source(self, vals):
+        conditions = [
+            vals.get("origin_exchange_record_id"),
+            "source_id" not in vals,
+            hasattr(self, "source_id"),
+        ]
```
simahawk:
```suggestion
        conditions = [
            "source_id" in self._fields,
            vals.get("origin_exchange_record_id"),
            "source_id" not in vals
        ]
```

edi_utm_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+from odoo import api, models
+
+
+class EDIExchangeConsumerMixin(models.AbstractModel):
+    _inherit = "edi.exchange.consumer.mixin"
+
+    def _edi_update_source(self, vals):
```
simahawk, line 17: docstring pls

edi_utm_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -0,0 +1,33 @@
...
+            vals.get("origin_exchange_record_id"),
+            "source_id" not in vals,
+            hasattr(self, "source_id"),
+        ]
+
+        if all(conditions):
+            utm_source_edi = self.env.ref(
+                "edi_utm_oca.utm_source_edi",
+                raise_if_not_found=False,
+            )
```
simahawk:
```suggestion
            utm_source_edi = self._edi_get_utm_source(vals)
```
to offer the hook for customization

simahawk (changes requested):

edi_utm_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -0,0 +1,39 @@
...
+    def _edi_get_utm_source(self):
+        return self.env.ref(
+            "edi_utm_oca.utm_source_edi",
+            raise_if_not_found=False,
+        )
+
+    def _edi_update_source(self, vals):
+        """Sets source_id to a predefined UTM source
+        if the record may have related EDI exchange records."""
+
```
simahawk:
```suggestion
        """Set UTM source for records created via EDI.
        """

```

edi_utm_oca/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1 @@
+* Chau Le <chaulb@trobz.com>
```
simahawk, line 2: can you list me here pls?

simahawk (approved):
Minor remarks

santostelmo (approved):

PR #105, opened 2024-07-30
[16.0][IMP] edi_oca: Split exchange error and traceback

HviorForgeFlow:
Having the exception error in the tree view can be very effective to identify error on several exchange records.

This PR aims to manage the exception message and the traceback in different fields in order let the user classify by error messages in the tree view and also take a look on the traceback on the form view.

CC @ForgeFlow

### DIFF
--- a/edi_oca/models/edi_backend.py
+++ b/edi_oca/models/edi_backend.py
@@ -20,7 +20,13 @@ from ..exceptions import EDIValidationError
 _logger = logging.getLogger(__name__)
 
 
-def _get_exception_msg():
+def _get_exception_msg(exc):
+    if hasattr(exc, "args") and isinstance(exc.args[0], str):
+        return exc.args[0]
+    return repr(exc)
+
+
+def _get_exception_traceback():
     buff = StringIO()
     traceback.print_exc(file=buff)
     traceback_txt = buff.getvalue()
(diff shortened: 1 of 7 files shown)


simahawk:
@HviorForgeFlow question: don't you have the traceback on the job?

HviorForgeFlow:
Yes I believe that it exists there, but there is no link directly. 

Error traceback is already in EDI exchange record, the only change in this PR is that as Traceback is usually big and from an user experience prespective someone asked me if I could put just the exception at tree level view to allow them to quick search common errors or understand better without having to enter to the form or look into the linked job.

simahawk:
> Yes I believe that it exists there, but there is no link directly.

We have now a btn to jump to the jobs linked to the record :wink: 


> Error traceback is already in EDI exchange record, the only change in this PR is that as Traceback is usually big and from an user experience prespective someone asked me if I could put just the exception at tree level view to allow them to quick search common errors or understand better without having to enter to the form or look into the linked job.

Exactly because it's big and very technical, I don't think it's a good idea to store it on the record as well.
Also, on jobs you can already search and group for the exception.

HviorForgeFlow:
But then since https://github.com/OCA/edi-framework/commit/76938cfc8ca41feb6349a6e04d8bd93a6c768e57 we are saving the traceback in the exchange records. Should we revert it and just add a optional column in the tree view?

simahawk:
Ah, good point! Well, I have no strong opinion. For sure, for how things are today is probably better to go ahead w/ this and separate them. Also, having an exchange on error is not normal and will be fixed manually or by an auto operation which means the tb want be stored for too long.

simahawk (approved):
Can you please backport to 14.0?

etobella (approved):

PR #109, opened 2024-09-25
[16.0][FIX] edi_oca: consider parent record when checking access rule

JordiMForgeFlow:
This fix prevents an error raised when you have child exchanges that do not have a specific model and res_id assigned.

When getting the related record of a child we are already looking at the parent if the exchange record does not have a specific model set (https://github.com/OCA/edi-framework/blob/d484f0dcbb7b1caca1aaae0f9a3c2b3510a714f5/edi_oca/models/edi_exchange_record.py#L210). The same should be done when checking the access.

### DIFF
--- a/edi_oca/models/edi_exchange_record.py
+++ b/edi_oca/models/edi_exchange_record.py
@@ -573,10 +573,15 @@ class EDIExchangeRecord(models.Model):
         for exc_rec in self.sudo():
             if not exc_rec.related_record_exists:
                 continue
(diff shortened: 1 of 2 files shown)


JordiMForgeFlow:
@etobella @simahawk could you review this one? :)

simahawk (commented):
LG... any chance to have a test?

JordiMForgeFlow:
@simahawk done :)

simahawk (approved):

PR #119, opened 2024-10-28
[17.0][MIG] edi_record_metadata_oca

etobella:
Migration to 17.0

Direct migration

@simahawk

PR #120, opened 2024-10-28
[17.0][MIG] edi_sale_oca: Migration and Split of code

etobella:
Module has been splitted in two modules:

1- Basic module with integration between edi_oca and sale (edi_sale_oca)
2- Module integrating edi_sale_oca and sale_order_import (edi_sale_import_oca)

It has been done this way in order to minimize dependencies on a base module. We could need edi_sale_oca but not all the functionalities provided by sale_order_import (it has a lot of dependencies)

Depends on:
- [ ] https://github.com/OCA/edi-framework/pull/119
- [ ] https://github.com/OCA/edi/pull/1003

@simahawk


simahawk:
@etobella I missed this. Can we please make a point on the many things going on at framework level? :)
Also, keep in mind that it is refactored here https://github.com/OCA/edi/pull/1067

PR #118, opened 2024-10-24
[17.0][FW] edi_oca: missing FW from 16.0

JordiMForgeFlow:
In the migration to V17 several commits were left behind in V16.

Depends on https://github.com/OCA/queue/pull/705

CC @etobella @simahawk

PR #122, opened 2024-12-02
[17.0][IMP] edi_oca: remove logs' warnings causes

SilvioC2C:
Remove usage of ``active_id`` from view

NB: there's another warning due to the fact that both fields `edi.exchange.type.rule.model` and `edi.exchange.type.rule.model_id` have the same string. However, that warning is already handled in https://github.com/OCA/edi-framework/pull/118.

### DIFF
--- a/edi_oca/views/edi_exchange_type_views.xml
+++ b/edi_oca/views/edi_exchange_type_views.xml
@@ -67,10 +67,8 @@
                             string="Model rules"
                             groups="edi_oca.group_edi_advanced_settings_manager"
                         >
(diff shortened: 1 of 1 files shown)


yankinmax (approved):

SilvioC2C:
NB: there's another warning due to the fact that both fields `edi.exchange.type.rule.model` and `edi.exchange.type.rule.model_id` have the same `string`. However, that warning is already handled in #118.

simahawk (approved):

PR #69, opened 2024-02-21
[17.0][MIG] edi_webservice_oca: Migration to 17.0

john-herholz-dt:

PR #123, opened 2024-12-02
[17.0][MIG] edi_webservice_oca: Migration to 17.0

SilvioC2C:
Supersedes #69


john-herholz-dt (approved):

yankinmax (approved):

gurneyalex (approved):

john-herholz-dt:
@SilvioC2C are you using this module already? I cannot get past the user error in components/send.py:

```
raise exceptions.UserError(
    _("`method` is required in `webservice` type settings.")
) from err
```

Currently there is - as far as I see it - no way that the component gets the webservice dict from the advanced exchange type settings.

SilvioC2C:
@john-herholz-dt yes, already in use.

I can share with you an example of the `.xml` file with the `edi.exchange.type` record's field `advanced_settings_edit`:
```
        <field name="advanced_settings_edit">
components:
    send:
        usage: webservice.send
        work_ctx:
            webservice:
                method: post
                kwargs:
                    url_params:
                        path: some/path
        </field>
```

EDIT: the result you get from this snippet when reaching the try/except block you quoted:
```
>>> self.ws_settings
{'method': 'post', 'kwargs': {'url_params': {'path': 'some/path'}}}
>>> method = self.ws_settings["method"].lower()
>>> method
'post'
```

john-herholz-dt:
Thanks that was it!

SilvioC2C:
You're welcome! I've been struggling with it myself, so no worries. Maybe I'll add it in the docs if I have time...

PR #126, opened 2024-12-06
[17.0][ADD] edi_project_oca: New module

SilvioC2C:


gurneyalex (approved):

edi_project_oca/models/project_project.py
```diff
@@ -0,0 +1,29 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import api, fields, models
+
+
+class ProjectProject(models.Model):
+    _name = "project.project"
+    _inherit = ["project.project", "edi.exchange.consumer.mixin"]
+
+    edi_disable_auto = fields.Boolean()
```
simahawk: no need to re-define this flag

edi_project_oca/models/project_project.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        projects = super().create(vals_list)
+        for project, vals in zip(projects, vals_list, strict=True):
+            project._event("on_project_create").notify(project, vals)
```
simahawk: no need to trigger these events. They are already available by default w/ component_event as `on_record_create` and `on_record_write`. 
What's the final goal for you?

edi_project_oca/models/project_task.py
```diff
@@ -0,0 +1,29 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import api, fields, models
+
+
+class ProjectTask(models.Model):
+    _name = "project.task"
+    _inherit = ["project.task", "edi.exchange.consumer.mixin"]
+
+    edi_disable_auto = fields.Boolean()
```
simahawk: same

edi_project_oca/models/project_task.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
```
simahawk: same

edi_project_oca/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,61 @@
+This module intends to create a base to be extended by local EDI rules
+for project management.
+
+In order to add customizations for projects, create a listener:
+
+```python
+from odoo.addons.component.core import Component
+
+
+class ProjectEventListenerExample(Component):
```
simahawk, line 10: I don't understand why you need these listeners

simahawk (changes requested):

edi_project_oca/models/project_project.py
SilvioC2C: I used `oca_account_edi` as example to create this module, and the field is re-declared there for `account.move`, so I did the same for consistency

simahawk: that's wrong unless you want to control the visibility by the state

edi_project_oca/views/project_project.xml
```diff
@@ -0,0 +1,61 @@
...
+                            <field name="exchange_record_count" />
+                        </span>
+                        <span class="o_stat_text">EDI</span>
+                    </div>
+                </button>
+            </xpath>
+            <xpath expr="//page[@name='settings']/group[last()]" position="inside">
+                <group
+                    name="edi_project_group"
+                    string="Electronic Data Interchange"
```
simahawk:
```suggestion
                    string="EDI"
```

edi_project_oca/views/project_task.xml
```diff
@@ -0,0 +1,42 @@
...
+                    <div class="o_form_field o_stat_info">
+                        <span class="o_stat_value">
+                            <field name="exchange_record_count" />
+                        </span>
+                        <span class="o_stat_text">EDI</span>
+                    </div>
+                </button>
+            </xpath>
+            <xpath expr="//page[@name='extra_info']/group[last()]" position="inside">
+                <group name="edi_project_group" string="Electronic Data Interchange">
```
simahawk:
```suggestion
                <group name="edi_project_group" string="EDI">
```
Q: wouldn't be better to have a specific page for EDI? 
Just asking...

SilvioC2C: Yep, can do that

edi_project_oca/tests/common.py
```diff
@@ -0,0 +1,34 @@
...
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+
+    @classmethod
+    def _build_and_add_component(cls, component_cls):
+        assert isinstance(component_cls, MetaComponent)
+        assert issubclass(component_cls, AbstractComponent)
+        component_cls._build_component(cls.comp_registry)
+        cls.comp_registry._cache.clear()
+
+    def _check_msg_in_log_output(self, msg: str, outputs: Iterable):
```
simahawk: kind of complicating things? :smile: 
See here how you can collect dummy events https://github.com/OCA/edi/blob/14.0/edi_oca/tests/fake_components.py#L10
Yet... se below...

edi_project_oca/tests/test_project_listener.py
```diff
@@ -0,0 +1,61 @@
...
+import logging
+
+from odoo.addons.component.core import Component
+
+from .common import TestEdiProjectOcaCommon
+
+_logger = logging.getLogger(__name__)
+
+
+class TestProjectListener(TestEdiProjectOcaCommon):
```
simahawk: you don't need all this testing... these are std events provided by component_event. You can simply drop them :wink:

simahawk (changes requested):

simahawk (approved):

PR #128, opened 2024-12-11
[17.0][MIG] edi_exchange_deduplicate_oca: Migration to 17.0

SilvioC2C:


edi_exchange_deduplicate_oca/models/edi_exchange_record.py
```diff
@@ -39,7 +39,7 @@ def create(self, vals_list):
 
     def _edi_get_duplicates(self, count=None):
```
Ricardoalso, line 40: Should we not set the boolean `False` for the count parameter instead of `None` ?

Ricardoalso (approved):
Small comment otherwise LGTM

edi_exchange_deduplicate_oca/models/edi_exchange_record.py
SilvioC2C: Well, we only use it as a check for deciding whether we're gonna use `search()` or `search_count()`, so as long as it can be evaluated as a boolean, it doesn't make much difference, any `False`-ish default value should be fine :smile:

Ricardoalso: Indeed but I find it a bit misleading and in the search method [back in V16,](https://github.com/odoo/odoo/blob/16.0/odoo/models.py#L1511C5-L1511C77) the default value was `False`

But this is just design :smile:

vvrossem:
Hello dear @OCA/edi-framework-maintainers, could you please reopen this PR and apply the "no stale" label?

yankinmax:
Hello dear @OCA/edi-framework-maintainers, could you please reopen this PR and apply the "no stale" label?

yankinmax:
Hello @ivantodorovich @jcoux @vvrossem can you pls take a look?

ivantodorovich (approved):

vvrossem (approved):

yankinmax:
Hello @OCA/edi-framework-maintainers can you take a look and possibly trigger merge?

Issue #127, opened 2024-12-10
edi_xml_oca: replace xmlschema with lxml to implement XML Schema validation


gurneyalex:
Issue

the  `edi_xml_oca` module uses `xmlschema` for the validation of XML documents. This is a pure Python library, and is can be really slow, especially with complex schemas, which are unfortunately numerous in the EDI world. 

We propose to replace the implementation with `lxml`, with is `libxml2` backed (or to use `lxml` if it is available, and fall back to `xmlschema` in case `lxml` is unavailable) to get a faster implementation.

simahawk:
hanks for taking time to write this down!
I had this in mind since a while, indeed.
That lib sucks perf wise.
Feel free to move on w/ that.
The only drawback is that that lib has been used to load data as well (likely just in some tests). To be checked.

simahawk:
FTR you have an example of validation here https://github.com/odoo/odoo/blob/834e8a8789d1d451f62ae1e841845baf7d3931a4/odoo/tools/xml_utils.py#L89 but it relies on having xsd as an attachment.

Ricardoalso:
I opted to use the [xmltodict library](https://pypi.org/project/xmltodict/) to convert the xml_content into a dictionary, replacing the previous approach that used the to_dict method from xmlschema. Additionally, @gurneyalex discovered [this Stack Overflow discussion](https://stackoverflow.com/questions/7684333/converting-xml-to-dictionary-using-elementtree) featuring an intriguing raw Python implementation. Do you have any preferences on this choice, @simahawk?

One open question remains: how should we handle the import and usage of this new dependency to maintain backward compatibility? Any thoughts or preferences, @simahawk?

gurneyalex:
closed by #129

PR #129, opened 2024-12-11
[16.0][IMP] edi_xml_oca: replace xmlschema by lxml

Ricardoalso:
Addressing the issue https://github.com/OCA/edi-framework/issues/127

### DIFF
--- a/edi_xml_oca/__manifest__.py
+++ b/edi_xml_oca/__manifest__.py
@@ -14,5 +14,5 @@
     "maintainers": ["simahawk"],
     "website": "https://github.com/OCA/edi-framework",
     "depends": ["edi_oca", "component"],
-    "external_dependencies": {"python": ["xmlschema"]},
+    "external_dependencies": {"python": ["xmltodict"]},
 }
--- a/edi_xml_oca/components/xml_handler.py
+++ b/edi_xml_oca/components/xml_handler.py
@@ -2,13 +2,12 @@
 # @author: Simone Orsi <simahawk@gmail.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
-import io
-from contextlib import closing
-
-import xmlschema
+import xmltodict
+from lxml import etree
 
 from odoo import modules
-from odoo.tools import DotDict
+from odoo.exceptions import UserError
+from odoo.tools.xml_utils import _check_with_xsd
 
 from odoo.addons.component.core import Component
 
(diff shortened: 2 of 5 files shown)


edi_xml_oca/__manifest__.py
```diff
@@ -14,5 +14,5 @@
     "maintainers": ["simahawk"],
     "website": "https://github.com/OCA/edi-framework",
     "depends": ["edi_oca", "component"],
-    "external_dependencies": {"python": ["xmlschema"]},
+    "external_dependencies": {"python": ["xmltodict", "lxml"]},
```
gurneyalex: I'm not sure that you need to declare lxml as a dependency, since it is a standard odoo dependency.

edi_xml_oca/components/xml_handler.py
```diff
@@ -65,15 +68,13 @@ def validate(self, xml_content, raise_on_fail=False):
...
             * error string if `raise_on_fail` is False
         """
+        xsd_name = self.schema_path
+        xml_content = (
+            xml_content.encode("utf-8") if isinstance(xml_content, str) else xml_content
+        )
         try:
-            return self.schema.validate(xml_content)
-        except self._validate_swallable_exceptions() as err:
+            validate_xml_from_attachment(self.env, xml_content, xsd_name)
```
simahawk: Are you sure this method is doing what we want? Have you checked it's implementation?

Ricardoalso: This public function validates the XML content with an XSD attachment; making a call to the private function `_check_with_xsd` that you pointed [here](https://github.com/OCA/edi-framework/issues/127#issuecomment-2531579658). Actually `validate_xml_from_attachment` acts as a wrapper to validate XML content against an XSD schema while handling exceptions gracefully.

The method implementation can be seen [here](https://github.com/odoo/odoo/blob/834e8a8789d1d451f62ae1e841845baf7d3931a4/odoo/tools/xml_utils.py#L315) 

Unless I'm missing something, yes I think the function does what we want :thinking:

simahawk: have you read the docstring of that method?
> If the XSD attachment cannot be found in database, skip validation without raising.

We do not have any attachment here. IMO we can use `_check_with_xsd` by passing directrly the stream of the XSD.
Then if we want to support attachments too, ok but this is not the case today.

Ricardoalso: Indeed  I missed the point that it was not being stored.  I refactored the code and added a new test to check if XML validation was done correctly

edi_xml_oca/components/xml_handler.py
```diff
@@ -28,52 +27,60 @@ def __init__(self, work_context):
...
-        :param file_obj: file obj of XML file
+        :param xml_string: str of XML file
         """
-        return DotDict(self.schema.to_dict(file_obj, **kw))
+        parsed_dict = xmltodict.parse(xml_string)
+        root_node = next(iter(parsed_dict))
+        return parsed_dict[root_node]
 
-    def parse_xml(self, file_content, **kw):
+    def parse_xml(self, file_content):
```
simahawk: are we sure we don't need these key args?

edi_xml_oca/static/description/index.html
```diff
@@ -8,10 +8,11 @@
 
 /*
 :Author: David Goodger (goodger@python.org)
-:Id: $Id: html4css1.css 8954 2022-01-20 10:10:25Z milde $
+:Id: $Id: html4css1.css 9511 2024-01-13 09:50:07Z milde $
```
simahawk: pls disable readme update w/ `export SKIP=oca-gen-addon-readme`

edi_xml_oca/components/xml_handler.py
Ricardoalso: It is indeed in the parameters of `xmltodict.parse` https://github.com/martinblech/xmltodict/blob/master/xmltodict.py#L184

Fixed

gurneyalex (approved):
Huge speedup!

simahawk (approved):

PR #131, opened 2024-12-19
[16.0] edi_storage_oca: Add migration Script

etobella:
Replaces #95 
Still pending to test it, but it is an idea.
@simahawk


etobella:
Tested and seems to work properly :smile:

luisDIXMIT (approved):
LGTM!

simahawk:
> Tested and seems to work properly 😄

Let's move on then! :)
Thank you!

/ocabot merge minor

PR #133, opened 2025-01-16
[18.0][MIG] edi_exchange_deduplicate_oca: Migration to 18.0

Ricardoalso:


wouitmil (approved):

yankinmax (approved):

edi_exchange_deduplicate_oca/models/edi_exchange_record.py
```diff
@@ -37,7 +37,7 @@ def create(self, vals_list):
                     obsoleted_records.edi_exchange_state = "obsolete"
         return records
 
-    def _edi_get_duplicates(self, count=None):
+    def _edi_get_duplicates(self, count=False):
```
dreispt, line 40: Why is this change needed?

Ricardoalso: Not needed but seems to be more accurate IMHO. As the parameter count is here used as a boolean flag to choose between two behaviors: calling either search_count or search

dreispt: Understood, that makes sense, I agree.

Ricardoalso:
superseded by https://github.com/OCA/edi-framework/pull/242 with the new edi_core_oca module

PR #136, opened 2025-01-23
[18.0][ADD] edi_exchange_template_oca_json

imlopes:


edi_exchange_template_oca_json/models/edi_exchange_template_output.py
```diff
@@ -0,0 +1,43 @@
...
+        result = self._render_json_values(exchange_record, **kw)
+        return result
+
+    def _post_process_output(self, output):
+        result = super()._post_process_output(output)
+        if self.generator == "json":
+            result = self._post_process_json_output(result)
+        return result
+
+    def _render_json_values(self, exchange_record, **kw):
```
simahawk, line 28: this module shouldn't rely on `jsonifier`.
You should be able to generate JSOn as you prefer.

The snippet can simply be something like
```
exporter = env.ref(...)
parser = exporter.get_json_parser()
result = record.jsonifier(parser, [one=True])
```
The role of this module is only to handle and validate the JSON result.

simahawk (changes requested):

simahawk:
Additionally, please, drop the odoo version from the commit. Thanks :)

wouitmil:
Hello, Italo, as I wasn't able to push on you're branch during your timeoff, I created this PR instead: https://github.com/OCA/edi-framework/pull/137

PR #132, opened 2025-01-15
[18.0][MIG] edi_exchange_template: Migration to 18.0

victorvermot:

PR #138, opened 2025-02-11
[18.0] edi_exchange_template_oca: declare tmpl to use explicitly

yankinmax:
Forward-port of:
- [x] https://github.com/OCA/edi/pull/1123


yankinmax:
@simahawk I need your help to fix the tests, I can't catch where I should remove `type_id` and specify explicitly tmpl

simahawk:
The failure comes from the deprecation warning but IMO we can drop this feature here in v18 as probably nobody is using it yet (we C2C migrated this module to v18 here https://github.com/OCA/edi-framework/pull/132)

simahawk:
@etobella @HviorForgeFlow @LoisRForgeFlow I'd like to get your feedback here :)

This module have been merged only 3 weeks ago by us and for us is not a problem to trash the old behavior.
If you are not aware of anyone relying on the old mechanism to lookup templates on v18, I'd like to trash the old deprecated behavior.

Of course, we can keep the migration script for who's passing from 17 to 18.

LoisRForgeFlow:
@simahawk Thanks for the headsup :1st_place_medal: , from my side you can rework in 18 as needed as long as you provide the needed migration scripts. We don't have any installation with this module in 18.

edi_exchange_template_oca/migrations/18.0.1.1.0/post-migrate.py
```diff
@@ -0,0 +1,60 @@
+# Copyright 2025 Camptocamp SA (http://www.camptocamp.com)
```
simahawk: @yankinmax  as we decided to drop the deprecated implementation, please change this. 
We should simply move this code to the pre-migrate step but we should create via SQL

* the column `output_template_id` col on the type
* the table `edi_exchange_template_type_rel` 

and set their values via SQL.
Keep in mind that the new fields won't be available at ORM level :wink: 

.Then you can trash the deprecated code + deprecated tests/

simahawk (changes requested):

edi_exchange_template_oca/migrations/18.0.1.1.0/post-migrate.py
yankinmax: @simahawk I updated this PR that way to be sure it won't break any build while we're handling the deprecated code.
I'll do the job in the coming weeks

edi_exchange_template_oca/migrations/18.0.1.1.0/pre-migrate.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2025 Camptocamp SA (http://www.camptocamp.com)
```
yankinmax: @simahawk I need your help here.
1. Do you have an idea how we can test the script?
2. Code review of deprecated feature.
Thanks

simahawk:
@yankinmax is this live?

yankinmax:
> @yankinmax is this live?

hello @simahawk 
I've written the migration script as you've suggested and asked you for a review.
What do you mean live? Is it used?
Yep, on some of the projects.

edi_exchange_template_oca/migrations/18.0.1.1.0/pre-migrate.py
```diff
@@ -0,0 +1,69 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from openupgradelib.openupgrade import logged_query
+
+from odoo.tools.sql import column_exists, create_column, table_exists
+
+
+def migrate(cr, version):
+    if not version:
+        return
+    if not table_exists(cr, "edi_exchange_template_type_rel"):
```
simahawk: @yankinmax if it exists we shouldn't do anything, no?

simahawk (changes requested):
Can you pls...
1. avoid updating the README?
2. rewrite the commit `fixup! edi_oca: fix test, because tmpl is set explicitly` -> it's not a fixup, it's an adaptation for v18

edi_exchange_template_oca/migrations/18.0.1.1.0/pre-migrate.py
yankinmax: `README.rst` is updated by pre-commit

simahawk (commented):
Thank you!

simahawk (approved):

simahawk:
build canceled, we have to update the version

edi_exchange_template_oca/migrations/18.0.1.1.0/pre-migrate.py
simahawk: the version should be 1.2.0, I fix it

simahawk:
> build canceled, we have to update the version

done

simahawk:
FTR I've just found out that this commit was not ported completely https://github.com/OCA/edi/commit/de46bac7cb25211a5cf8fa2912d1cffca9e41cea#diff-1530f22233009319365b6d21e8b6ed7a7dd53c6120cb22ddf5d32edfb6e93a9f

in particular the part to use the explicit template and the removal of the fallback

yankinmax:
> in particular the part to use the explicit template and the removal of the fallback

I'll try to find some time to check it again, sorry for that

simahawk:
I fixed it here https://github.com/OCA/edi-framework/pull/182

PR #142, opened 2025-02-22
[18.0][IMP] edi_exchange_template_oca: add json support

wouitmil:
This PR aims to merge the module edi_exchange_template_oca_json in edi_exchange_template_oca

If a template is of type json. payload key is waited in the result. result['payload'] contains then the generated json


simahawk (changes requested):
This change should be done in the base module.

Side note: do not add the odoo version in the commit https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

wouitmil:
> This change should be done in the base module.
> 
> Side note: do not add the odoo version in the commit https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

In fact, shouldn’t all these modules be merged into the base module? JSON is quite common, and it doesn’t require any dependencies.

simahawk:
@wouitmil I've cleanup commits and changed the lookup for the post processor.

simahawk (approved):

PR #147, opened 2025-03-18
[18.0][ADD] move edi_oca abstract parts into edi_core_oca

acsonefho:
Based on https://github.com/OCA/edi/issues/1131

- Move models from `edi_oca` into `edi_core_oca`;
- Move views from `edi_oca` into `edi_core_oca`;
- Move security from `edi_oca` into `edi_core_oca`;
- Move cron from `edi_oca` into `edi_core_oca`;
- Move templates from `edi_oca` into `edi_core_oca`;
- Move job config (channels etc) from `edi_oca` into `edi_core_oca`;
- Move wizard from `edi_oca` into `edi_core_oca`;
- Move utils and exception from `edi_oca` into `edi_core_oca`;
- Move static (JS) from `edi_oca` into `edi_core_oca`;
- Move data from `edi_oca` into `edi_core_oca`;
- Move demo from `edi_oca` into `edi_core_oca`;
- Move abstract tests parts from `edi_oca` into `edi_core_oca`;
- Drop abstract tests parts from `edi_oca` and move them into `edi_core_oca` (keep specific tests with `Components` in `edi_oca` );
- Update XMLID used everywhere;
- Update syntax of write with `odoo.Command.xxx()`;
- Adapt readme parts of the new `edi_core_oca` (should be improved).

Todo:
- Should I increase the version number as a major change?


edi_account_oca/views/res_partner.xml
```diff
@@ -7,7 +7,7 @@
         <field name="model">res.partner</field>
         <field name="inherit_id" ref="account.view_partner_property_form" />
         <field name="arch" type="xml">
-            <!-- TODO: Move this conf inside `edi` page from edi_oca. -->
+            <!-- TODO: Move this conf inside `edi` page from edi_core_oca. -->
             <group name="accounting_entries" position="after">
```
simahawk, line 11: I'd drop this change in this PR and move it to its own. The page is already available on the partner.

edi_oca/models/edi_backend.py
```diff
@@ -291,411 +148,71 @@ def _validate_data(self, exchange_record, value=None, **kw):
...
-                }
-            )
-            if (
-                state == "input_processed_error"
-                and old_state != "input_processed_error"
-            ):
-                exchange_record._notify_error("process_ko")
-            elif state == "input_processed":
-                exchange_record._notify_done()
-        exchange_record.notify_action_complete("process", message=message)
```
simahawk, line 529: @acsonefho can you please restore the calls to `notify_action_complete`? This is a common interface to trigger stuff when something happens on a record. The only thing to re-implement would be `_trigger_edi_event`. 
There, you can leave it empty in edi_core and edi_oca will only override this to trigger events.

simahawk (changes requested):
Nice job, thank you. So far, the only thing that I would change is the notification part. See my comment.

edi_oca/models/edi_backend.py
acsonefho: Thank you for your review! I did some fix related to this PR into this one https://github.com/OCA/edi-framework/pull/149
That's the place where I notice that think kind of feature doesn't work anymore.
So I have to move some changes done in the other PR here to be clean.

simahawk: can you cherry-pick them here so we can move on w/ this part?

simahawk: @acsonefho are you going to move on soon w/ this?
We have pipeline of migration PR that are kept on hold for this.
Would be nice if we can integrate this right away, otherwise it will have to wait.
Just let me know. Thanks!

acsonefho: Hi @simahawk !
Unfortunately I don't have time this week then I'm on holidays until July :/

acsonefho:
I close it as the job is done and merged here: https://github.com/OCA/edi-framework/pull/192

PR #148, opened 2025-03-19
[16.0][IMP] edi_oca set job priority from exchange type

santostelmo:
Allows you to set the job priority configured in exchange type


simahawk (approved):
LG. Thank you!

PR #135, opened 2025-01-16
[18.0] [MIG] edi_storage_oca: Migration to 18.0

Ricardoalso:

PR #149, opened 2025-03-19
[18.0][ADD] move edi_storage_oca abstract parts into edi_storage_core_oca

acsonefho:
Based on:
- https://github.com/OCA/edi-framework/pull/147
- https://github.com/OCA/edi-framework/pull/135


simahawk:
@acsonefho what are you trying to achieve? What's the purpose to remove components here?

edi_storage_core_oca/abstracts/base.py
```diff
@@ -0,0 +1,71 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+import logging
+from pathlib import PurePath
+
+from .. import utils
+
+_logger = logging.getLogger(__file__)
+
+
+class EDIStorageMixin:
```
simahawk, line 13: To be honest I don't like these mixin classes.
I'd rather move the methods to a specific python class.
Eg:

```
class StorageManager:

    def __init__(self, storage):
        self.storage = storage

    # put here all the common methods from components


# on the edi backend

def _storage_get_manager(self):
    return StorageManager(self.storage_id)

```

Then it's up to you how you implement the storage handling and you won't need `edi.backend.fs.storage`.

edi_storage_core_oca/abstracts/listener.py
```diff
@@ -0,0 +1,76 @@
...
+
+import functools
+import os
+from pathlib import PurePath
+
+from .. import utils
+
+
+class EdiStorageListenerAbstract:
+    def _move_file(self, storage, from_dir_str, to_dir_str, filename):
```
simahawk, line 12: these methods - except `on_*` -  go to the storage manager as well

simahawk (changes requested):

acsonefho:
> @acsonefho what are you trying to achieve? What's the purpose to remove components here?

IMO Components don't provide any benefits (for this case).
It's easier to work with AbstractModel.

I don't want to load some Components module just to have a simple EDI.

simahawk:
> > @acsonefho what are you trying to achieve? What's the purpose to remove components here?
> 
> IMO Components don't provide any benefits (for this case). It's easier to work with AbstractModel.
> 
> I don't want to load some Components module just to have a simple EDI.

Well, you have to do all this work because you don't want to use components :stuck_out_tongue_winking_eye: 
You don't have to look at components in the ctx of the storage, you have to look at them in the ctx of the framework.
Anyway, pls check my comments above and we can maybe simplify this thing.

PR #150, opened 2025-03-20
[16.0][FIX] edi_storage_oca: do not silently ignore OSError/FileNotFoundError

santostelmo:
Depends on https://github.com/OCA/queue/pull/674

if an OSError is raised, there can be different cases but it's likely a transient issue (e.g. connection error):

if we ignore it as currently:
Job won't fail
the exchange record will be incorrectly marked as "input_received" but with no data, which is not what we want
so we don't ignore it anymore:
Job will fail;
the record state will stay as "input_pending" (as exception won't be catched in [edi.backend:exchange_receive](https://github.com/OCA/edi-framework/blob/16.0/edi_oca/models/edi_backend.py#L537))
and as such, another action_exchange_receive Job will be created next time scheduled action "EDI exchange check input sync" runs, to try again to receive it, hopefully successfully this time
a RetryableJobError will be raised in order to retry it
if it sill can't succeed after max_retries, record state will be changed to "input_receive_error"
if a FileNotFoundError is raised, it's not a transient issue, so record should be marked as "input_receive_error":

to do so we don't ignore it anymore:
it will get catched in [edi.backend:exchange_receive](https://github.com/OCA/edi-framework/blob/16.0/edi_oca/models/edi_backend.py#L537) as a ["swallable"](https://github.com/OCA/edi-framework/blob/16.0/edi_oca/models/edi_backend.py#L548) (Final) error
record state will be changed to "input_receive_error"

### DIFF
--- a/edi_oca/models/edi_exchange_record.py
+++ b/edi_oca/models/edi_exchange_record.py
@@ -332,6 +332,42 @@ class EDIExchangeRecord(models.Model):
         self.ensure_one()
         return self.backend_id.exchange_receive(self)
 
+    def action_exchange_send_on_max_retries_reached(self):
+        self.ensure_one()
+        self.write(
+            {
+                "edi_exchange_state": "output_error_on_send",
+                "exchange_error": self.env.context.get("exc_info", "Error on Send"),
+                "exchanged_on": fields.Datetime.now(),
+            }
+        )
+        self._notify_error("send_ko")
+        return None
+
+    def action_exchange_process_on_max_retries_reached(self):
+        self.ensure_one()
+        self.write(
+            {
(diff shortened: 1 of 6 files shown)


edi_oca/models/edi_exchange_record.py
```diff
@@ -332,6 +332,42 @@ def action_exchange_receive(self):
         self.ensure_one()
         return self.backend_id.exchange_receive(self)
 
+    def action_exchange_send_on_max_retries_reached(self):
```
simahawk, line 335: These methods should be renamed based on the new implementation on qjob side.
They should be private BTW.

simahawk (changes requested):

PR #151, opened 2025-04-25
[16.0][FIX] edit_storage_oca get_input_filenames return filename

santostelmo:
File names should not be prefixed by relative path


gurneyalex (approved):

simahawk:
> File names should not be prefixed by relative path

this looks like a breaking change

PR #154, opened 2025-05-12
[18.0][MIG] edi_backend_partner_oca: Migration to 18.0

AlexPForgeFlow:
Migration to 18.0


JordiMForgeFlow (approved):

LoisRForgeFlow (approved):
:+1:

AaronHForgeFlow (approved):
Code ok

JordiMForgeFlow:
CC @simahawk @etobella

etobella:
Silly question. Doesn't it clash with the idea of edi conf?

JordiMForgeFlow:
@etobella we use the link more as static information, to know that a specific partner has a dedicated backend configured. Not sure if the edi conf forces you to configure some flow or if it can be used for the same purpose, we will test :)

JordiMForgeFlow:
@etobella I have reviewed and you are right, it makes sense to embrace the EDI Conf and with that we can already get the relation of the backend with the partner.

etobella:
Should we close it then?

PR #157, opened 2025-05-15
[FIX] edi_oca: Fix get_view

etobella:
In some weird cases, the fields were not loaded.

They way to fix it was to set it this way.


etobella:
@simahawk your comments would be awesome in this topic :pray:

LoisRForgeFlow (approved):
Cannot reproduce but I do not anticipate any problem that could be generated by the change

etobella:
I think that the problem can be raised if the field is declared on the view for some reason

etobella:
/ocabot merge patch

It is the same logic applied in base_tier_validation

PR #158, opened 2025-05-26
[18.0][FIX] edi-oca: error in component key

victorvermot:
With the current implementation we had issues where t-key was undefined several times in the loop.
In a foreach we cannot have duplicated  t-key which means that we ran into tracebacks when opening views (only in debug mode).
I modified the t-key used to be more robust and avoid this kind of issues

### DIFF
--- a/edi_oca/static/src/xml/widget_edi.xml
+++ b/edi_oca/static/src/xml/widget_edi.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <templates id="template" xml:space="preserve">
     <div t-name="edi_oca.EdiConfigurationWidget">
-        <t t-foreach="info" t-as="rule_id" t-key="rule_index">
+        <t t-foreach="info" t-as="rule_id" t-key="rule_id_index">
             <t t-set="rule" t-value="info[rule_id]" />
             <t t-if="rule.form">
                 <t t-set="btn" t-value="rule.form.btn" />


edi_oca/static/src/xml/widget_edi.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <templates id="template" xml:space="preserve">
     <div t-name="edi_oca.EdiConfigurationWidget">
-        <t t-foreach="info" t-as="rule_id" t-key="rule_index">
```
Ricardoalso, line 4: Also done in https://github.com/odoo/enterprise/blob/18.0/spreadsheet_edition/static/src/bundle/list/side_panels/edit_list_sorting_section/edit_list_sorting_section.xml#L15

Ricardoalso (approved):
LGTM

cyrilmanuel (approved):

yankinmax (approved):

simahawk (commented):
LG. Thanks

edi_oca/readme/CONTRIBUTORS.md
```diff
@@ -1,3 +1,4 @@
 - Simone Orsi \<<simahawk@gmail.com>\>
 - Enric Tobella \<<etobella@creublanca.es>\>
 - Manuel Regidor \<<manuel.regidor@sygel.es>\>
+- Victor Vermot \<<victorvermot@gmail.com>\>
```
simahawk: FTR Normally we don't add contributors for very small changes ;)

victorvermot: Sorry, was not sure if I had to do it, I'll remove it ;)

simahawk:
@victorvermot sorry, I didn't notice `edi-oca` instead of `edi_oca` in the commit msg :sweat_smile: 
Can you please fix it?

PR #167, opened 2025-06-04
[16.0][IMP] edi_oca: relate multiple records with a single exchange record

JordiMForgeFlow:
In our projects we have faced different scenarios where we have needed to link more than one Odoo record to an exchange record. For example, in a warehouse integration, you could be exchanging a Product Masterdata file including multiple products, and you want them all to be linked to the exchange record.

Currently the EDI framework only allows to link one Odoo record to an exchange record. Using the current code, a solution could be to use the ability to link exchange records in a parent/child relation. However, in the use case presented you will only sent one file, so it is a bit weird to have child exchange records that will simply do nothing.

My goal is to try to find a solution that fits the current code and allows covering these type of use cases. As a generic approach, I am proposing this POC where we could simply add a model to relate multiple records (being from different models) with a single exchange record.

Looking forward to get your opinions!

Supersedes https://github.com/OCA/edi-framework/pull/42


JordiMForgeFlow:
@simahawk @etobella better later than never. We have been using this POC for some time now and it may be interesting to get it included :)

etobella:
We need a migration script for this.

IMO, it is not a problem.

JordiMForgeFlow:
@etobella @simahawk ready for review :)

edi_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -59,6 +63,13 @@ def _compute_edi_config(self):
             record.edi_config = config
             record.edi_has_form_config = any([x.get("form") for x in config.values()])
 
+    @api.depends("exchange_related_record_ids")
+    def _compute_exchange_record_ids(self):
+        for rec in self:
+            rec.exchange_record_ids = rec.mapped(
+                "exchange_related_record_ids.exchange_record_id"
+            )
```
simahawk:
```suggestion
            rec.exchange_record_ids = rec.exchange_related_record_ids.exchange_record_id
```

edi_oca/models/edi_exchange_record.py
```diff
@@ -187,6 +200,20 @@ def _compute_ack_expected(self):
         for rec in self:
             rec.ack_expected = bool(self.type_id.ack_type_id)
 
+    @api.depends("related_record_ids")
+    def _compute_record(self):
+        """
+        Main related record is computed as the first related record created
+        for the exchange.
+        """
+        for rec in self:
```
simahawk, line 250: This could be expensive when you have tons of records.
IMO we should do a select for all records at once to get the 1st one right away.

edi_oca/models/edi_exchange_record.py
```diff
@@ -259,6 +286,10 @@ def name_get(self):
     def create(self, vals_list):
         for vals in vals_list:
             vals["identifier"] = self._get_identifier()
+            if vals.get("res_id") and vals.get("model"):
```
simahawk: what about an inverse method on res_id?

edi_oca/models/edi_exchange_record.py
```diff
@@ -383,7 +414,15 @@ def action_open_related_record(self):
         return self.record.get_formview_action()
 
     def _set_related_record(self, odoo_record):
-        self.sudo().update({"model": odoo_record._name, "res_id": odoo_record.id})
+        self._set_related_records(odoo_record)
```
simahawk: I would deprecate `_set_related_record` in favor of `_set_related_records`.

edi_oca/models/edi_exchange_record.py
```diff
@@ -599,7 +626,37 @@ def check_access_rule(self, operation):
 
     def write(self, vals):
         self.check_access_rule("write")
-        return super().write(vals)
+        if "res_id" not in vals and "model" not in vals:
+            return super().write(vals)
+        # Updating main record information
+        res_id_updated = bool("res_id" in vals)
```
```suggestion
        res_id_updated = "res_id" in vals
```
it's already a boolean

edi_oca/views/edi_exchange_record_views.xml
```diff
@@ -197,6 +197,25 @@
...
+                            <field name="related_record_ids" nolabel="1" colspan="4">
+                              <tree create="false" delete="false" editable="bottom">
+                                <field name="model" />
+                                <field name="res_id" />
+                                <button
+                                        name="action_open_related_record"
+                                        type="object"
+                                        icon="fa-arrow-right"
+                                        string="Open Record"
+                                        attrs="{'invisible': ['|', ('model', '=', False), ('res_id', '=', False)]}"
```
simahawk: The invisible check is useless IMO: they cannot be empty.

edi_oca/models/edi_exchange_record.py
```diff
@@ -599,7 +626,37 @@ def check_access_rule(self, operation):
...
+            vals["related_record_ids"] = related_records_command_list
+            return super().write(vals)
+        if res_id_updated and res_id and model_updated and model:
+            # We can make bulk update
+            related_records_command_list.append(
+                Command.create({"res_id": res_id, "model": model})
+            )
+            vals["related_record_ids"] = related_records_command_list
+            return super().write(vals)
+        # Only updating res_id or model. We need to go one by one
```
simahawk: I'm not sure to understand why this is needed. Can you describe what is the expected behavior in general?
Eg: if you touch `res_id` or `model` you keep the old relation while replacing the main one?

IMO to keep things simple, you should always add a relation, either w/ an inverse on res_id or via `_set_related_records`. This way you don't need this complex machinery in write.

Then if you want to remove a record, it should be explicit, via a new method `_drop_related_records`.

WDYT?

simahawk (changes requested):
Nice work! :pray:

edi_oca/models/edi_exchange_record.py
JordiMForgeFlow: how would you handle it? should we adapt the already existing '_set_related_record' to handle multiple odoo_record?

edi_oca/models/edi_exchange_record.py
JordiMForgeFlow: yes, the inverse is cleaner. However, it does not work correctly in the 'create' unless the res_id and model are readonly=False (I have also tried with the force_save on the view).

Another option would be to leave them readonly=False and making them readonly in the views

edi_oca/models/edi_exchange_record.py
JordiMForgeFlow: The goal is to make sure that custom logics that are currently working only with the main res_id and model keep working as expected. I was thinking on cases where any of these fields are changed, then the related records should also be updated to remove the old ones and add the new one. I agree that an inverse will be cleaner for this case.

edi_oca/models/edi_exchange_record.py
JordiMForgeFlow: I was thinking on doing a read_group to get the results but I am facing a problem due to res_id being an int. If I do the following:

self.env["edi.exchange.related.record"].read_group(
    domain=[("exchange_record_id", "in", self.ids)],
    fields=["exchange_record_id", "res_id", "model"],
    groupby=["exchange_record_id"],
)

it adds up the value of all the res_id of the related records.

Should we do a explicit SQL query?

edi_oca/models/edi_exchange_record.py
simahawk: I'd go for the last option

edi_oca/models/edi_exchange_record.py
simahawk: Have you tried w/ `limit=1` + `orderby="id"` ?
In any case we don't mind having sql here.

edi_oca/models/edi_exchange_record.py
simahawk: That's already the case w/ your change. However, thinking of it again, we should probably make our minds work w/ `_set_related_record` being able to work on multiple records and forget about `_set_related_records` :smile:

JordiMForgeFlow:
@simahawk comments attended 🙏🏼

JordiMForgeFlow:
Hi @simahawk 

Could we move on with this one?

edi_oca/README.rst
```diff
@@ -1,7 +1,3 @@
-.. image:: https://odoo-community.org/readme-banner-image
```
simahawk: tip: `export SKIP=oca-gen-addon-readme` to not touch readme files :wink:

edi_oca/models/edi_exchange_record.py
```diff
@@ -187,6 +202,66 @@ def _compute_ack_expected(self):
...
+        for rec in self:
+            rec.res_id = data.get(rec.id, {}).get("res_id", False)
+            rec.model = data.get(rec.id, {}).get("model", False)
+
+    def _inverse_record(self):
+        """
+        Create related record when writing the main record for the exchange.
+        """
+        for rec in self:
+            if not rec.res_id or not rec.model or rec.related_record_ids:
```
simahawk: so you assume that if you have `related_record_ids` valued you don't need to check if this record is already there?

simahawk (commented):
LG overall. One minor remark

edi_oca/README.rst
JordiMForgeFlow: oh, didn't know about this one. Should I revert the changes on the README file or is it fine to leave them?

edi_oca/models/edi_exchange_record.py
JordiMForgeFlow: you are right, better to check if it is there in case it needs to be added 👍🏼

JordiMForgeFlow:
Hi @simahawk 

How is this one looking now? :)

GuillemCForgeFlow (approved):
useful feature, code LGTM 👍🏿

edi_oca/README.rst
simahawk: they just make noise and - worse part - when you cherry pick commits it's easy to get conflicts. Can you roll this back pls?

simahawk (commented):
LG overall

JordiMForgeFlow:
@simahawk done :)

simahawk (approved):

simahawk:
Great work! Thanks! 
/ocabot merge minor

JordiMForgeFlow:
@simahawk I will prepare the FW :)

GuillemCForgeFlow:
fw for 17 already done here: https://github.com/OCA/edi-framework/pull/189, I will rebase once this is merged 👍🏿

PR #168, opened 2025-06-04
[16.0] edi_oca misc fwd ports from edi/14.0

simahawk:


etobella:
@simahawk TEsts are failing

simahawk:
Needs more attention, I'll get back to this. Stays in draft for now.

simahawk:
I've moved commits not related to edi_oca to https://github.com/OCA/edi-framework/pull/169

simahawk:
@etobella could you have a look pls?

etobella:
@simahawk I cannot do now this kind of review (we are talking of 1000 lines of code and 15 different commits). If all of them are reviewed and merged in previous version we can probably proceed. 

It would be easier if at each PR merged, a forward - port commit is created to all branches. I know it is harder for developer/contributor, but it is easier for reviewer

simahawk:
> @simahawk I cannot do now this kind of review (we are talking of 1000 lines of code and 15 different commits). If all of them are reviewed and merged in previous version we can probably proceed.
> 
> It would be easier if at each PR merged, a forward - port commit is created to all branches. I know it is harder for developer/contributor, but it is easier for reviewer

@etobella thanks for your feedback!. Yeah, I know but I really didn't have time before this week. Now I do, so I'll fwd port to 18 right away. In fact, I don't mind that much about 16 but I wanted to keep 16 and 18 in line as far as possible before moving to 18. 

One important point: commits are all atomic: easy to review 1 by 1 :wink: 

Anyway, what's here is already live on 14 since a lot of time. There's no change in core methods or logic. The big part is the addition of edi.configuration which does not impact existing flows.

In any case I'll be working on EDI for the next few weeks so I'll be available if there's anything to fix :wink:

wouitmil (approved):

sebalix (approved):
LG, no issue with fake model loader anymore then with `at_install`?

simahawk:
> LG, no issue with fake model loader anymore then with `at_install`?

Nope. In fact was the lack of that same decorator to cause the issue (it was already applied to the other 2 test cases using model loader).

PR #180, opened 2025-07-09
[18.0][MIG] edi_purchase_oca: Migration to 18.0

P-H-Phuc:
Migration `edi_purchase_oca` to version 18.0


edi_purchase_oca/models/purchase_order.py
```diff
@@ -0,0 +1,21 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import models
+
+
+class PurchaseOrder(models.Model):
+    _name = "purchase.order"
+    _inherit = ["purchase.order", "edi.exchange.consumer.mixin"]
+
+    def button_confirm(self):
```
simahawk: this part is obsolete as we  have specific events now or even the possibility to not use components.

simahawk: this part is missing https://github.com/OCA/edi/pull/1098

simahawk: As said, this part is obsolete as we have the listeners that will trigger the edi.configuration machinery -> we don't need this file anymore.

edi_purchase_oca/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,14 @@
+This module intends to create a base to be extended by local edi rules
+for purchase.
+
+In order to add a new integration, you need to create a listener:
+
+``` python
+class MyEventListener(Component):
```
simahawk: this is outdated. The way to configure this is just to assign the edi.configuration on the partner form

simahawk (changes requested):

simahawk:
replaced by https://github.com/OCA/edi-framework/pull/276

PR #182, opened 2025-07-15
[18.0] edi_exchange_template: complete specific tmpl lookup

simahawk:
Fix the behavior according to the feature ported from v14
and drop the deprecated lookup by code.

Missing work from https://github.com/OCA/edi-framework/pull/138


yankinmax (approved):
Thanks @simahawk

PR #184, opened 2025-07-15
[18.0] edi_endpoint: fix migration to v18

simahawk:
A new route parameter has been introduced and for existing endpoints to work OOTB we must resync.

Companion https://github.com/OCA/web-api/pull/108


simahawk:
Tested on real project.
/ocabot merge patch

PR #186, opened 2025-07-17
[18.0] edi: fix edi.configuration.action_view_partners

simahawk:
Fix Uncaught Promise > View types not defined tree found in act_window action undefined.


simahawk:
Fast tracking as it's broken now.
/ocabot merge patch

PR #192, opened 2025-08-08
[18.0][IMP] edi_*: Remove dependency on component and queue and set it on a different modules

etobella:
~~This module is not intended to be merged on 18, but it is my suggestion for v19.~~
The intention is to merge the module in 18, but make it compatible with existent solutions (for this reason we added hooks and regenerated xml_ids)

The idea is to split the dependancies of edi_oca to avoid this extra dependancies that might make it harder to handle. Also, this way it  might be easier to have `edi_oca` before `queue_job`and `component_event`.

@simahawk your comments will be welcomed

At the end I will split the commits properly between modules and so on, but right now it makes sense to have everything at the same point

TODO:

- [x] Set component behavior on `edi_component_oca`
- [x] Set queue behavior on `edi_queue_oca`


etobella:
The issue is unrelated, as edi-framework is failing right now when installing account. Seems a problem with some library. I will check it if possible.

etobella:
Failure comes from the fact that

https://raw.githubusercontent.com/OCA/OCB/18.0/requirements.txt
is not the same than https://github.com/OCA/OCB/blob/18.0/requirements.txt

it is quite weird, because if we try to get the raw value directly from github, it redirects to https://raw.githubusercontent.com/OCA/OCB/refs/heads/18.0/requirements.txt and it fails.

I think it is time to ask @sbidoul if he has any idea about it.

etobella:
I think this might fix it:

https://github.com/OCA/oca-ci/pull/102

By the change of requirements, the stdnum library was not correct and everything started to fail (not only this repo, other repos are affected)

sbidoul:
This looks like a GitHub bug indeed.

sbidoul:
Side note: I have long wanted to have an event mechanism that is independent of component.

account_connector_listener/models/account_move.py
```diff
@@ -0,0 +1,30 @@
+# Copyright 2020 Dixmit
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
```
simahawk: I don't think we need such changes anymore. The framework already triggers events for state changes if you assign the consumer mixin (see edi_sale, edi_purchase, edi_stock, some PR are still pending) + configure `edi.configuration` records.

edi_oca/models/edi_backend.py
```diff
@@ -12,10 +12,9 @@
 
 from odoo import exceptions, fields, models
 
-from odoo.addons.component.exception import NoComponentError
 from odoo.addons.queue_job.exception import RetryableJobError
 
-from ..exceptions import EDIValidationError
+from ..exceptions import EDINotImplementedException, EDIValidationError
```
```suggestion
from ..exceptions import EDINotImplementedError, EDIValidationError
```

edi_oca/models/edi_backend.py
```diff
@@ -65,107 +63,7 @@ class EDIBackend(models.Model):
...
-        candidates = [record_conf["usage"]] if record_conf else []
-        candidates += [
-            base_usage,
-        ]
-        return candidates
-
-    def _get_component_conf_for_record(self, exchange_record, key):
-        settings = exchange_record.type_id.get_settings()
-        return settings.get("components", {}).get(key, {})
+    execution_model = fields.Char()
```
simahawk: If I get your proposal right, this should only stay on the type. 
It's only the type that states how an exchange should work.

edi_oca/models/edi_backend.py
```diff
@@ -736,3 +646,22 @@ def _is_valid_edi_action(self, action, raise_if_not=False):
 
     def _failed_output_check_send_msg(self):
         return "Nothing to do. Likely already sent."
+
+    def _get_execution_function(self, exchange_record, action):
```
simahawk: IMO only `_get_exec_handler` should raise `EDINotImplementError`.
Then in the component module you have to override only this and not all the methods.
Plus, you can simply check if `exchange_record.type_id.exec_handler_model` is valued.

FTR I had something in my mind since a while, above all to simplify the UI, to declare the handler on the type and make it mandatory. In the component module it can be used to look for the component.

However, this implementation, makes harder to decouple the actions.
Eg: you don't want to have the generate handler to be the same handler of the send as they should remain agnostic.

This means that we should probably have 4 fields, one per action handler.

simahawk (changes requested):
> Side note: I have long wanted to have an event mechanism that is independent of component.

Poor `component`, nobody loves you :sob: 

@etobella I'd like to discuss these changes on a call :wink: 

FTR There was a draft w/ a different approach here already https://github.com/OCA/edi-framework/pull/147
(Personally I prefer the changes here in #192)

edi_oca/models/edi_backend.py
etobella: Sure, why not, I was just making a proposal

edi_oca/models/edi_backend.py
etobella: I like your proposal. I will do it in a few weeks

etobella:
Well, components add an extra layer that might be too complicated sometimes. Also this PR removes the queue job dependency.

I would prefer to keep module names as we defined them here, but it is not blocking.

sbidoul:
cc/ @acsonefho

edi_oca/models/edi_backend.py
etobella: I did something for this using the config settings. Can you check it?

edi_oca/models/edi_backend.py
etobella: Done!

account_connector_listener/models/account_move.py
etobella: Well, I just did it in order to avoid confussion. In this case we can remove it later on the migration. In my case I will not use it anymore :wink:

edi_oca/models/edi_backend.py
etobella: Also, I renamed the function name in order to avoid future confusions

edi_endpoint_oca/__manifest__.py
```diff
@@ -12,7 +12,7 @@
     "license": "LGPL-3",
     "website": "https://github.com/OCA/edi-framework",
     "author": "Camptocamp,Odoo Community Association (OCA)",
-    "depends": ["base_edi", "edi_oca", "endpoint"],
+    "depends": ["base_edi", "edi_component_oca", "endpoint"],
```
simahawk: I don't think this is required.

edi_exchange_template_oca/README.rst
```diff
@@ -1,7 +1,3 @@
-.. image:: https://odoo-community.org/readme-banner-image
```
simahawk, line 1: can you set `SKIP=oca-gen-addon-readme` to avoid readme updates?

edi_oca/models/edi_backend.py
```diff
@@ -634,10 +540,11 @@ def _exchange_receive_check(self, exchange_record):
...
     def _exchange_receive(self, exchange_record):
-        component = self._get_component(exchange_record, "receive")
-        if component:
-            return component.receive()
-        raise NotImplementedError()
+        exchange_function = self._get_exec_handler(exchange_record, "receive")
+        if exchange_function:
+            ctx = self._get_record_env_ctx(exchange_record, "receive")
+            return exchange_function(exchange_record.with_context(**ctx))
+        raise EDINotImplementedError("No handler for `receive`")
```
simahawk: to simplify: `_get_exec_handler` should raise if not found

edi_oca/models/edi_backend.py
```diff
@@ -669,7 +576,7 @@ def _check_input_exchange_sync(self, record_ids=None, **kw):
             len(pending_process_records),
         )
         for rec in pending_process_records:
-            rec.with_delay().action_exchange_process()
+            rec._check_input_exchange_sync_process()
```
simahawk: another layer to cleanup: crons. I'm not sure we need them by default nowadays.

edi_oca/models/edi_exchange_type.py
```diff
@@ -88,6 +84,7 @@ class EDIExchangeType(models.Model):
         comodel_name="edi.exchange.type",
         compute="_compute_ack_for_type_ids",
     )
+    execution_model = fields.Char()
```
simahawk:
```suggestion
    execution_handler_model_id = fields.Many2one()
```
this could be a selection filtered by inheriting mixin and/or attribute.

Another option is that we rely on pure python classes (We could have a dotted path to the python class here).
It could work this way (thinking out loud):
```
# base_handler

class BaseHandler:
  __slots__ = ("exchange_record",)
  def __init__(self, exc_record):
      self.exchange_record =  exc_record

  @property
  def record(self):
    return self.exchange_record.record

# my_handler.py

from odoo.addons.edi_oca.handlers import BaseHandler

class MyHandler(BaseHandler):

  def generate(self):
      return "ok"

# edi_exchange_type.py

from ..my_handler import MyHandler

class ExcType(..):
  _inherit = "edi.exc.type"

   execution_handler = fields.Selection(selection_add=[(MyHandler._klass_dotted_path(), "My handler")])
```

Pros
- clean and simple
- no pollution of odoo models
- no need to migrate models if they change
- contextual object (instead of having abstract models that are heavier and that cannot be bound to a specific record on init)

Cons
- you cannot extend an handler as an odoo model, you have to define your own handler if you want that (but that would make it more explicit)

etobella:
Well, after talking with @simahawk in a meeting we have decided the following:

- [x] We will create a new module called edi_core_oca where the logic of edi_oca will be passed to.
- [x] edi_oca will depende don `edi_core_oca`, `edi_component_oca` and `edi_queue_oca`. It will contain all the common and other classes linked in order to avoid issues with existent modules
- [x] We will create abstract classes for handlers (one for each kind) and we will allow to select them with ir.model and correct filters.

We evaluated to use classes for the handlers, but it follows the odoo way to use abstract classes (with their pros and cons).

With this in mind, we think that we can merge the changes in v18. I will work on it on the following days and if no problems are risen, we will proceed with the merge.

ValentinVinagre:
> We will create a new module called edi_core_oca where the logic of edi_oca will be passed to.
> 

The edi_oca module that it will contain?

etobella:
> The edi_oca module that it will contain?

Not too much, no real code, just references. It will be done this way in order to avoid issues with other modules that might exist, like ediversa 😉

The idea is to avoid changes in dependant modules

etobella:
@simahawk Should I move the commit history? This way we will not loose the changes done (I assume that `edi_oca` will not be migrated to 19). This can be easily achieved now :wink:

ValentinVinagre:
> @simahawk Should I move the commit history? This way we will not loose the changes done (I assume that `edi_oca` will not be migrated to 19). This can be easily achieved now :wink: 

IMO. Yes, it should be transferred

simahawk:
> @simahawk Should I move the commit history? This way we will not loose the changes done (I assume that `edi_oca` will not be migrated to 19). This can be easily achieved now 😉

If you mean "move to edi_core", yes.
Maybe the easiest is to just do `git mv edi_oca edi_core_oca` and recreate `edi_oca` from scratch?

etobella:
Don't worry, I will handle that when we are ready for the merge :wink:

acsonefho:
@etobella I already open an issue about this on https://github.com/OCA/edi/issues/1131 and I did the work on https://github.com/OCA/edi-framework/pull/147 (+ others PR based on this).
It's used in production (and of course, it's working like a charm).
My PRs are not clean enough so I plan to clean them during OCA days.

The best is maybe to wait and improve mines?

etobella:
@acsonefho I read your PR (after doing my first approach, sorry). Your PR is interesting but mine includes some extra features (split of queue) and it has been tested with existent EDI works and works as expected.

also, in this case, XMLIDs are ported to the right module but it has been refreshed in in edi_oca in order to minimize existent modules issues.

Also, I would like to merge this next week (a month before OCA Days). Because we need that in several OCA repos (l10n-spain, document_quick_access...)

Obviously, if you have something that it is missing in my branch, I can check your module and import it here (even maintaining autorship for this things if you want). No problem from my side.

etobella:
Before merge, I will rewrite history of the PR so we have the right commit everywhere (and I will push the history of edi_oca to edi_core_oca) :smile:

etobella:
@simahawk Any change on the current modules might give problems on the merge. We might need to be fast on this. Can I kindly ask for your review? (I will do all the commit stuff at the end)

etobella:
@simahawk :green_apple: :green_apple: 

as you can see in edi_stock_oca, the only change needed was on the tests, as we were making a patch on them.

I made a test with the migration https://github.com/OCA/edi-ediversa/pull/1 and everything worked fine (with the old version and the new version) I did not the migration as there was needed a patch in taxes, otherwise, I would push that :wink:

etobella:
About rewriting commit history. I can do that also in v19 migration in order to simplify this PR. As now it has minimal dependancies, I can do the migration in the OCA Days with the rewritting of commits to avoid the struggle now. WDYT?

edi_endpoint_oca/__manifest__.py
etobella: It is done this way to allow the usage of edi_endpoint_oca without edi_oca.

edi_exchange_template_oca/README.rst
etobella: Sure!

edi_oca/models/edi_backend.py
etobella: Well, we leave it for just one case, when we don't have the quick exec. It might be removed in the future for sure.

rousseldenis:
@etobella @simahawk 

@acsonefho has invested much time on this following correctly OCA conventions, making an issue to discuss about that before doing the job.

@etobella Why don't you have commented #147 instead ?

rousseldenis:
It should be great if you have a call as suggested by @simahawk all together no ?

etobella:
@rousseldenis we had the call a week ago where we agreed in a lot of things to do.

I didn't answered because I didn't notice the issue or PR. I was not expecting this kind of change, sorry. That kind of things might happen in Opensource projects. Actually, the same happened with the creation of the edi framework.

However my approach makes an extra split (removing queue job) and includes handlers based on models following PSC agreement on that topic. Also, as I said in other comments, I am open to include some of your changes if there is something missing. Also, it would be interesting to merge it fast.

Another thing, I made it in a way to minimize issues with dependent modules, so it is possible that the dependent modules need no changes in views and so on, as xmlids have been duplicated. Was it done in a similar way?

simahawk:
@etobella I'll try to find some time today to review this final state :sweat_smile: 

@rousseldenis me and Enric are in line on what we want to achieve.

We had a call and we discussed many aspects, including the simplification of the configuration (which goes beyond the original intent but it goes in the right direction to simplify things w/o "just" removing components).
A short summary has been reported here https://github.com/OCA/edi-framework/pull/192#issuecomment-3207156791

I've also pointed out that I prefer the approach of this PR https://github.com/OCA/edi-framework/pull/192#pullrequestreview-3105984021 over the one by @acsonefho (which BTW did not include all the changes as some of them were split in the storage PR, that I didn't like as well, sorry).

Finally, Enric offered to move fast on this and to make it smooth to transition to this new approach w/o breaking existing installations and I agreed on doing this in v18 only because of this (and because I can test it on a migration project soon).

So, here we go. Nothing personal :innocent:

edi_core_oca/hooks.py
```diff
@@ -0,0 +1,61 @@
...
+        UPDATE ir_model_data
+        SET module = 'edi_core_oca'
+        WHERE module = 'edi_oca'
+        """
+    )
+    openupgrade.rename_xmlids(
+        env.cr,
+        [
+            (
+                "edi_core_oca.edi_conf_trigger_record_create",
```
simahawk: this looks wrong because the original name is `edi_oca` no?

edi_core_oca/__manifest__.py
```diff
@@ -0,0 +1,53 @@
+# Copyright 2020 ACSONE
```
simahawk, line 1: The 1st commit looks wrong as it moves edi_oca inside core instead of renaming it. This is fixed only in the 2nd commit but hidden.

edi_core_oca/hooks.py
```diff
@@ -0,0 +1,61 @@
...
+            ("name", "=", "edi_oca"),
+            ("state", "in", ["installed", "to upgrade", "to install"]),
+        ]
+    ):
+        for module in ["edi_core_oca", "edi_queue_oca", "edi_component_oca"]:
+            for data in env["ir.model.data"].search([("module", "=", module)]):
+                if not env["ir.model.data"].search_count(
+                    [("module", "=", "edi_oca"), ("name", "=", data.name)]
+                ):
+                    data.copy({"module": "edi_oca", "noupdate": True}).name = data.name
```
simahawk, line 37: I'm not sure to understand what you are doing here :thinking:

edi_core_oca/models/edi_backend.py
```diff
@@ -409,16 +312,9 @@ def exchange_generate_send(self, recordset, skip_generate=False, skip_send=False
...
-                job1 = rec.delayable().action_exchange_generate()
-                # Chain send job.
-                # Raise prio to max to send the record out as fast as possible.
-                job1.on_done(rec.delayable(priority=0).action_exchange_send())
-                job1.delay()
-            elif skip_send:
-                rec.with_delay().action_exchange_generate()
-            elif not skip_send:
-                rec.with_delay(priority=0).action_exchange_send()
+            rec._exchange_generate_send(
```
simahawk: Hmm this is making a ping-pong between the backend and the record IMO.
AFAIU it's to allow the override of the method on the record just to call `with_delay`, right?

If that's the case, we can simply remove the `with_delay` here and use [_patch_job_auto_delay](https://github.com/OCA/queue/blob/18.0/queue_job/models/base.py#L145) in the queue module to automatically delay the methods.

`action_generate_send` instead, can be fully overridden to chain jobs.

edi_core_oca/models/ir_model.py
```diff
@@ -0,0 +1,46 @@
...
+
+    is_edi_generator = fields.Boolean()
+    is_edi_input_validator = fields.Boolean()
+    is_edi_output_validator = fields.Boolean()
+    is_edi_sender = fields.Boolean()
+    is_edi_receiver = fields.Boolean()
+    is_edi_processor = fields.Boolean()
+    is_edi_checker = fields.Boolean()
+
+    def _reflect_model_params(self, model):
```
simahawk, line 18: Do we really need all these fields?
If it is to ease the m2o filtering can't we simply have a string field w/ the action?
And, the action can be set by default as an attribute on the base class (eg: `_edi_framework_action = 'generate'`) and here you store in in `edi_framework_handler` chat field.
If empty, is not an handler, if valued, you know its purpose. WDYT?

edi_oca/tests/test_consumer_mixin.py
```diff
@@ -10,14 +10,13 @@
...
 
-from odoo.tests import Form, tagged
+from odoo.tests import Form
 
 from .common import EDIBackendCommonTestCase
 
 
 # This clashes w/ some setup (eg: run tests w/ pytest when edi_storage is installed)
 # If you still want to run `edi` tests w/ pytest when this happens, set this env var.
-@tagged("at_install", "-post_install")
```
simahawk, line 20: why this change?

edi_core_oca/tests/fake_models.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class EdiExchangeConsumerTest(models.Model):
+    _name = "edi.exchange.consumer.test"
+    _inherit = ["edi.exchange.consumer.mixin"]
+    _description = "Model used only for test"
+
+    name = fields.Char()
+    edi_config_ids = fields.Many2many(
+        string="EDI Purchase Config Ids",
```
simahawk:
```suggestion
        string="EDI test Config Ids",
```

edi_core_oca/views/edi_exchange_type_views.xml
```diff
@@ -66,10 +64,47 @@
                         </group>
                     </group>
                     <notebook>
+                        <page name="Execution models" string="Execution models">
```
simahawk:
```suggestion
                        <page name="exec_handlers" string="Execution handlers">
```

edi_core_oca/views/edi_exchange_type_views.xml
```diff
@@ -66,10 +64,47 @@
...
+                            <group name="input_models" invisible="direction != 'input'">
+                                <field
+                                    name="generate_model_id"
+                                    options="{'no_create': True}"
```
simahawk, line 61: shall we make these fields required at least in the view?

edi_core_oca/views/edi_exchange_type_views.xml
```diff
@@ -66,10 +64,47 @@
                         </group>
                     </group>
                     <notebook>
+                        <page name="Execution models" string="Execution models">
+                            <group name="input_models" invisible="direction != 'input'">
+                                <field
+                                    name="generate_model_id"
```
simahawk: shouldn't be `process_*`?

edi_core_oca/views/edi_exchange_type_views.xml
```diff
@@ -66,10 +64,47 @@
...
+                                    options="{'no_create': True}"
+                                />
+                                <field
+                                    name="send_model_id"
```
simahawk: receive?

edi_endpoint_oca/__manifest__.py
simahawk: but we don't use components here, isn't it?

edi_oca/__manifest__.py
```diff
@@ -1,56 +1,16 @@
...
-    "name": "EDI",
-    "summary": """
-    Define backends, exchange types, exchange records,
-    basic automation and views for handling EDI exchanges.
-    """,
-    "version": "18.0.1.3.2",
-    "website": "https://github.com/OCA/edi-framework",
-    "development_status": "Beta",
+    "name": "Edi Oca",
+    "summary": """Integrate all EDI modules together""",
```
simahawk, line 6: I would state this is now a transition / proxy module that probably will be discarded, because bla bla bla.
And I would not touch the authorship.

simahawk (changes requested):

edi_core_oca/hooks.py
etobella: but it is moved to edi core before that :wink:

edi_core_oca/hooks.py
etobella: We are recreating the xmlids and assign them on edi_core. We set noupdate so it is not deleted on updates. This is done because this way, if we have a module that refers to a view of edi_oca (for example), it will work as expected :wink:

edi_core_oca/views/edi_exchange_type_views.xml
etobella: If we do them required, we need to unset them when we install edi_component_oca or edi_exchange_template_oca :thinking:

edi_core_oca/__manifest__.py
etobella: It was done this way because it is simpler to rebase this way :wink:  That is part of the job previous to merge.

edi_oca/tests/test_consumer_mixin.py
etobella: Because by default it is at_install and not post_install and we removed the post_isntall -at_install condition in edi_core_oca (left as it was in edi_oca)

edi_core_oca/models/ir_model.py
etobella: If we do it this way, we need to generate one model for each action, this way, we can reuse the same model, check:

https://github.com/OCA/l10n-spain/pull/4352/files#diff-152f2b726090062611c591e9bfef9c05d603d22d23cd0a59cdaa2a500883a628R7

edi_core_oca/models/edi_backend.py
etobella: Well, it is not so easy in this case.... if you check the function, on edi_core_oca, it is generated directly, but in the second case, it is done with a chained set of jobs :thinking:

acsonefho:
@etobella 

> That kind of things might happen in Opensource projects

Yes. Except that you have possibility to propose some improvements into existing PR.

> @acsonefho I read your PR (after doing my first approach, sorry). Your PR is interesting but mine includes some extra features (split of queue) and it has been tested with existent EDI works and works as expected.


So my full contribution will be totally dropped just because you didn't check if an issue/PR is already open and you don't want to contribute in it.

> Obviously, if you have something that it is missing in my branch, I can check your module and import it here (even maintaining autorship for this things if you want). No problem from my side.

More than that, I should improve your PR? :smile: :cherries: 

**So to summary a little bit:**
- We have the same goal (split to remove `Component`) + you have others improvements to do (drop dependency with `queue_job` etc) :muscle: ;
- You didn't check if an issue is open (or maybe too late) https://github.com/OCA/edi/issues/1131;
- You didn't check if another PR was open (or maybe too late) https://github.com/OCA/edi-framework/pull/147;
- When I notice you about that, you decide to completely ignore my work/contribution https://github.com/OCA/edi-framework/pull/192#issuecomment-3209738545;
- My customer paid for my working hours and I have to wait yours to completely rebase it and re-invoice this customer (on something who is used in production and who is working well; what can I say to him?).

Is it correctly summarized?

etobella:
@acsonefho 

First of all, I’m sorry that this situation happened to you. I really understand the frustration. The same thing actually happened to me 5 years ago, with Acsone and the same module.

I created the first EDI module on 19 of OCtober of 2020 (v13) based on the `l10n_es_facturae` (Spanish Electronic invoicing communication system) that we had been using for 3-4 years inside OCA in the spanish localization. A month later, Simone (under Acsone requirements) came with a different approach and similar goal but based on Components. I had two options, enforce my view or accept his as it was more mature. I accepted the second one. I also lost a lot of hours on this.

- https://github.com/OCA/edi/pull/243
- https://github.com/OCA/edi/pull/248

I am exposing this, because I understand your frustration. In any case, I am happy that we accepted the change done by Simone as I think it was the right approach back then. This kind of things happen in OpenSource projects.

In the case we have in our hands, I made a more complete approach:

- I have splitted all external modules in new modules, simplifying everything.
- I have tested it with an external module that it worked with edi_oca previously and kept working after the changes, with no changes in the module. I made the test using https://github.com/OCA/edi-ediversa/pull/2. Just notice that the module will fail with your approach as you removed xmlids on edi_oca, which breaks backward compatibility.

Looking at the situation around your PR:

- You started in June, and Simone gave feedback, but the work stalled for various reasons.
- Simone has already invested significant time in reviewing and adapting around your changes.
- The current proposal doesn’t yet include a smooth transition path, which will be important for acceptance in v18.

We (the maintainers) discussed this together, and from a technical and maintenance perspective, the consensus is to adopt my PR as the base. I will then carefully integrate the improvements from your branch, keeping module names and preserving authorship wherever possible.

This should give us a solid foundation for v18 and beyond, while ensuring your contributions are included and acknowledged.

WDYT?

simahawk:
Thank you Enric for the summary. I share this same vision.

@acsonefho One honest remark about this as it sounds a bit weird or - at least - not realizing that is not just about "your work".

> My customer paid for my working hours and I have to wait yours to completely rebase it and re-invoice this customer 
(on something who is used in production and who is working well; what can I say to him?).

My customers paid my working hours to review your work that I haven't planned at all and I kept all other framework PRs (needed for our customers) blocked to give you time to finish it until I realized you were off for holidays (and beyond) :cry:
I did not complain. I understand that's life and I wanted to help.  

Technically  speaking, your work would never be accepted in v18 at the current state because it's not complete (eg: no migrations): you would have to do exactly what Enric did here (and it's the only reason why I accepted to do it in v18 while his initial proposal was to go for v19).

Hence, you can simply tell to your customer "someone else improved and finished my work but I have to adapt few things to integrate it in the project". That's simple and true and it's the same thing that I will do :wink: 

That's OSS, pure and simple, in a collaborative community :trophy: 
We should all thanks each other IMO :relaxed:

acsonefho:
> @acsonefho
> 
> In the case we have in our hands, I made a more complete approach:
> 
> * I have splitted all external modules in new modules, simplifying everything.

I totally agree.  I did exactly the same.

> * I have tested it with an external module that it worked with edi_oca previously and kept working after the changes, with no changes in the module. I made the test using [[DO NOT MERGE][18.0][MIG] edi_ediversa_oca edi-ediversa#2](https://github.com/OCA/edi-ediversa/pull/2). Just notice that the module will fail with your approach as you removed xmlids on edi_oca, which breaks backward compatibility.

I agree too, that's a point that I probably have to fix. But not a reason to completely drop my work.
> 
> Looking at the situation around your PR:
> 
> * You started in June, and Simone gave feedback, but the work stalled for various reasons.

I open the issue https://github.com/OCA/edi/issues/1131 18th February and I started the dev 18th March https://github.com/OCA/edi-framework/pull/147 (+ others PR based on this one).
I was on holidays during last feedback and I didn't find time yet to take in consideration the review of Simone. And whatever this feedback, the minimum is to ask the author if we can start another PR from scratch or continue the work (by cloning the PR or anything else).

> * Simone has already invested significant time in reviewing and adapting around your changes.

And? When someone review and need some fixes, it's drop?

> * The current proposal doesn’t yet include a smooth transition path, which will be important for acceptance in v18.
What do you mean by "smooth transition path"?
> 
> We (the maintainers) discussed this together, and from a technical and maintenance perspective, the consensus is to adopt my PR as the base. I will then carefully integrate the improvements from your branch, keeping module names and preserving authorship wherever possible.

So you're saying that maintainers are ignoring open issue and existing PR and then make some consensus? Hum, _Do what I say, not what I do_
As contributor (and more as maintainer) you should have cloned my PR and improve it.

> 
> This should give us a solid foundation for v18 and beyond, while ensuring your contributions are included and acknowledged.
> 
> WDYT?

acsonefho:
> Thank you Enric for the summary. I share this same vision.
> 
> @acsonefho One honest remark about this as it sounds a bit weird or - at least - not realizing that is not just about "your work".
> 
By _my work_, I mean the time spent to split/fix/improve etc on PRs. Saying _my_ doesn't exclude any others contributors. :pray: 

> > My customer paid for my working hours and I have to wait yours to completely rebase it and re-invoice this customer
> > (on something who is used in production and who is working well; what can I say to him?).
> 
> My customers paid my working hours to review your work that I haven't planned at all and I kept all other framework PRs (needed for our customers) blocked to give you time to finish it until I realized you were off for holidays (and beyond) 😢 I did not complain. I understand that's life and I wanted to help.

I don't really see the point of this comment.

> 
> Technically speaking, your work would never be accepted in v18 at the current state because it's not complete (eg: no migrations): you would have to do exactly what Enric did here (and it's the only reason why I accepted to do it in v18 while his initial proposal was to go for v19).

You never mention that in your review. https://github.com/OCA/edi-framework/pull/147#pullrequestreview-2891958532

> 
> Hence, you can simply tell to your customer "someone else improved and finished my work but I have to adapt few things to integrate it in the project". That's simple and true and it's the same thing that I will do 😉
> 
> That's OSS, pure and simple, in a collaborative community 🏆 We should all thanks each other IMO ☺️

IMO we don't have the same definition of _collaborative_.

etobella:
@acsonefho 

> I totally agree. I did exactly the same.

No, you didn't do that:

- You created two modules, edi_core_oca(Without components) and edi_oca (With components).
- You didn't provide a way to handle edi different than components
- edi_core_oca had no sense without edi_oca, so the split was not filling any real needs.
- You had several months to handle that, You did nothing in that period. Your last commit is from 19th of March, Simone provided a review and you did nothing in 5 months... If I had found your PR and issue, I would have seen that there was no changes in the 4 months and I would have done the same I did.
- Your solution might break existent solutions.
- There were no migration scripts

> So you're saying that maintainers are ignoring open issue and existing PR and then make some consensus? Hum, Do what I say, not what I do
> As contributor (and more as maintainer) you should have cloned my PR and improve it.

Actually that is not true. I provided examples... It depends on the maturity of the solution, needs and real evolution. In this case, my option is more mature, complete and evolved completly independent. For that reason we are not keeping the history and leaving it open to include the improvements from yours to mine.

I think you are asking me to keep your commits to increase your ego or providing some justification in front of your customer. For me that is not enough reason to do that, as it is just rewriting history for your own benefit.

Also, leaving a PR dead for 4 months doesn't enforce me to keep your PR. You might have lost interest, as the Simone's review was 4 months old. Also, my work was done as a PoC for 19, not 18, but after talking with Simone, we decided to move it to 18 if we were able to ensure that everything keeps working.

> IMO we don't have the same definition of collaborative.

It might be. For me collaborative means to work together and decide for the biggest goal, not my own decisions. As I have done several times by finding the best approachs for all parts (as I did when I decided to join on Acsone/Simone's work in the past).

In this particular case, my goal is to provide the best technical solution for everyone and I think my solution has a better approach and scope.

simahawk:
Guys, please, let's keep this discussion constructive and to the point. To me the point is technical. **From now on I will ignore any comment that goes in the direction of personal feelings** "I've done this, I've done that, you don't/didn't, etc)".

@acsonefho can you please tell us what prevents you to integrate this work in your project?
So far you only complained about the modus operandi and you did not provide any technical argument about anything in the content of this PR.

Is there anything that is missing? Anything that is incompatible? Anything that... name yours pls.

Enric offered to cherry-pick changes from your PR if needed, as well as maintain the ownership or co-autorship of the main changes. If you can answer the questions above, we can move on.

Thanks for your help :pray:

acsonefho:
> @acsonefho
> 
> > I totally agree. I did exactly the same.
> 
> No, you didn't do that:
> 
> * You created two modules, edi_core_oca(Without components) and edi_oca (With components).

Yes

> * You didn't provide a way to handle edi different than components

I think you didn't really read/use/test my PR at all. But of course you know better than me what I did and what is actually runs in production.

> * edi_core_oca had no sense without edi_oca, so the split was not filling any real needs.

It's very strange because in production, I don't use `edi_oca` and the EDI is working well. I use only these PRs.
And of course, `Component` is not installed neither.

> * You had several months to handle that, You did nothing in that period. Your last commit is from 19th of March, Simone provided a review and you did nothing in 5 months... If I had found your PR and issue, I would have seen that there was no changes in the 4 months and I would have done the same I did.

I forgot that you were next to me when I work on that part and Github is lying when I see my last commit 30th May.
As maintainer, you probably didn't notice that I split into separated PR (± one per module).

> * Your solution might break existent solutions.

Yes you probably right. I try to avoid that kind of issue but maybe it should be improved.

> * There were no migration scripts

Yes but you can put a comment and add it. It's not a problem

> 
> > So you're saying that maintainers are ignoring open issue and existing PR and then make some consensus? Hum, Do what I say, not what I do
> > As contributor (and more as maintainer) you should have cloned my PR and improve it.
> 
> Actually that is not true. I provided examples... It depends on the maturity of the solution, needs and real evolution. In this case, my option is more mature, complete and evolved completly independent. For that reason we are not keeping the history and leaving it open to include the improvements from yours to mine.
> 
_my opinion is more mature_: Ok, I don't have nothing to say about that :smile: 

> I think you are asking me to keep your commits to increase your ego or providing some justification in front of your customer. For me that is not enough reason to do that, as it is just rewriting history for your own benefit.
> 

My ego? Seriously? That becomes personal. With that sentence, I'm asking about your role of maintainer. It's rude.

> Also, leaving a PR dead for 4 months doesn't enforce me to keep your PR. You might have lost interest, as the Simone's review was 4 months old. Also, my work was done as a PoC for 19, not 18, but after talking with Simone, we decided to move it to 18 if we were able to ensure that everything keeps working.
> 

He reviews it in June and we're in Augustus. Math says ±2 months.
So if I correctly follow you this procedure in this PR https://github.com/OCA/server-backend/pull/376 (of re-use previous commits done here https://github.com/OCA/server-backend/pull/338 ) was not correct?

> > IMO we don't have the same definition of collaborative.
> 
> It might be. For me collaborative means to work together and decide for the biggest goal, not my own decisions. As I have done several times by finding the best approachs for all parts (as I did when I decided to join on Acsone/Simone's work in the past).
> 

As you said, _together_ :smile: .
I'm not saying that your solution is not correct or anything else (that's not the point). I'm saying that you should reviews/improve/take care of my work before make a (solo-?) consensus about that you will merge your PR and drop others works (mine in this case). It's the case here or anything else in OCA.

> In this particular case, my goal is to provide the best technical solution for everyone and I think my solution has a better approach and scope.

acsonefho:
> Guys, please, let's keep this discussion constructive and to the point. To me the point is technical. **From now on I will ignore any comment that goes in the direction of personal feelings** "I've done this, I've done that, you don't/didn't, etc)".
> 
> @acsonefho can you please tell us what prevents you to integrate this work in your project? So far you only complained about the modus operandi and you did not provide any technical argument about anything in the content of this PR.
> 
> Is there anything that is missing? Anything that is incompatible? Anything that... name yours pls.
> 
> Enric offered to cherry-pick changes from your PR if needed, as well as maintain the ownership or co-autorship of the main changes. If you can answer the questions above, we can move on.
> 
> Thanks for your help 🙏

@simahawk Go ahead with that PR :pray:
It seems that my contribution is only for ego. I won't be part of that

etobella:
@acsonefho

I’d like to apologize for the tone of my last message. Reading it again, I see that it came across as too harsh and personal, which was not my intention. I let my own frustration get in the way of expressing things constructively, and I’m sorry for that.

As maintainers, our role is to balance different contributions and perspectives, and I should have kept the discussion more focused on that shared responsibility rather than making it feel personal.

We are humans and we fail sometimes, however, we need to be aware of that and fix our wrongs if possible. In this case, it is important to recognize my own failure and ask for your forgiveness.

edi_core_oca/models/ir_model.py
simahawk: Looks like it makes sense for verticalizations. Let's keep it as is :+1: 
I would add a comment here on why we need this stuff.

edi_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -408,20 +412,3 @@ def _edi_send_via_email(
         # Send the email
         composer.send_mail()
         return True
-
```
simahawk, line 420: I would be nice if could support executing edi.conf even w/o components.
There idea would be to call these methods:

* `_edi_on_create`
* `_edi_on_write`
* `_edi_on_before_state_change`
* `_edi_on_state_change`

and a `_edi_get_edi_config` to return the right recordset for edi.config for the current model.
Then edi_sale|stock|purchase etc can override this method to return the right field
or we can even rely on `_edi_config_field` attribute that can be taken automatically.

To verify: we might have to handle the exec from this API + events from component that could lead to executing configs twice. To be checked.

edi_exchange_template_oca/__manifest__.py
```diff
@@ -11,7 +11,7 @@
     "author": "ACSONE,Camptocamp,Odoo Community Association (OCA)",
     "maintainers": ["simahawk"],
     "website": "https://github.com/OCA/edi-framework",
-    "depends": ["edi_oca", "component"],
+    "depends": ["edi_core_oca"],
```
simahawk, line 14: It would be nice for the long term, to eliminate the override on the backend and replace it with:
* an handler for generate that use the selected tmpl
* and the tmpl field will be shown only if you select this specific generate handler

edi_core_oca/models/edi_exchange_record.py
```diff
@@ -599,58 +587,20 @@ def write(self, vals):
...
-        # Drop ID otherwise the context will be loaded from the action's record
-        action.pop("id")
-        return action
+    def _check_input_exchange_sync_receive(self):
+        self.action_exchange_receive()
+
+    def _check_input_exchange_sync_process(self):
+        self.action_exchange_process()
+
+    def _exchange_generate_send(self, skip_generate=False, skip_send=False):
```
simahawk: as we discussed we can move back this method to the backend and simply have a hook for the chaining of the jobs.

edi_core_oca/models/edi_exchange_type.py
```diff
@@ -88,6 +84,23 @@ class EDIExchangeType(models.Model):
...
+    )
+    process_model_id = fields.Many2one(
+        "ir.model", domain=[("is_edi_processor", "=", True)]
+    )
+    check_model_id = fields.Many2one("ir.model", domain=[("is_edi_checker", "=", True)])
+    generate_model_id = fields.Many2one(
+        "ir.model", domain=[("is_edi_generator", "=", True)]
+    )
+    send_model_id = fields.Many2one("ir.model", domain=[("is_edi_sender", "=", True)])
+    input_validate_model_id = fields.Many2one(
```
simahawk, line 102: we should check if we mention anywhere that the action is "validate" whereas now we define it as "input_validate" and "output_validate"... it's mostly for doc reasons.

edi_oca/utils.py
```diff
@@ -1,25 +1,3 @@
...
-
-
-def exchange_record_job_identity_exact(job_):
-    hasher = identity_exact_hasher(job_)
-    # Include files checksum
-    hasher.update(
-        str(sorted(job_.recordset.mapped("exchange_filechecksum"))).encode("utf-8")
-    )
-    return hasher.hexdigest()
+from odoo.addons.edi_core_oca.utils import get_checksum, normalize_string
```
simahawk, line 3:
```suggestion
from odoo.addons.edi_core_oca.utils import get_checksum, normalize_string
try:
    from odoo.addons.edi_component_oca.utils import exchange_record_job_identity_exact
except:
    _logger.info("deprecated....")
    bla bla
```
and if we log we should log for all the imports? But in v19 we'll trash this so.... whatever

edi_exchange_template_oca/__manifest__.py
etobella: sure, I can make it once it is merged (just to avoid a lot of new changes :wink:

edi_core_oca/models/edi_exchange_type.py
etobella: I rephrased the documentation with the new changes.

edi_core_oca/models/edi_exchange_record.py
etobella: :+1:

etobella:
@simahawk All has been changed as we talked about. As you told me, the first commits was messy and needed a lot to be fixed (the move was not correctly done). Sorry for the misunderstanding on that specific topic.

LoisRForgeFlow:
Hi @etobella @simahawk 

FWIW, I've tested this in a real staging environment, and after update, everything kept working without any additional changes. So, at least, it seems this does not break anything.

My 2 cents.

simahawk:
Thanks for testing @LoisRForgeFlow ! On my side I've been sick last week and I'm struggling to find time to do it. I'll try hard this week :sweat_smile:

etobella:
@simahawk This should be :green_apple: :green_apple: :green_apple: again.

I am sure that this week in Belgium, you will be able to review it and proceed with merge :pray:

simahawk:
@etobella I was finally able to test this on a real project: it works like a charm :tada: 
Do you still have to do any change?

etobella:
From my side we can proceed

etobella:
@simahawk With your approval we can merge :pray:

simahawk:
@etobella I'm a bit concerned about the UI. Now we have

image

In this case, using components we don't care about exec handlers.
How could we improve this? Shall we add a new field like `exec_engine` where we can choose `component` or `model` and hide/show accordingly.

An alternative, could be to:
* rename the 1st tab to "Execution"
* move advanced settings field to the bottom of the 1st tab

WDYT? Any other option?

edi_account_core_oca/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,2 @@
+This module intends to create a base to be extended by local edi rules
+for accounting.
```
simahawk, line 2: the icon of this module is outdated, non blocking

edi_component_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2025 Dixmit
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+
+from odoo import models
+
+
+class EdiExchangeConsumerMixin(models.AbstractModel):
+    _inherit = "edi.exchange.consumer.mixin"
+
+    def _manual_notify_edi_generation(self, exchange_record):
```
simahawk, line 11:
```suggestion
    def _edi_create_manual_notify(self, exchange_record):
```

edi_component_oca/readme/CONFIGURE.md
```diff
@@ -0,0 +1,16 @@
...
+The component used on edi must inherit from:
+
+- `edi.component.input.mixin` for processing and implement the process function
+- `edi.component.receive.mixin` for reception and implement the receive function
+- `edi.component.output.mixin` for generation and implement the generate function
+- `edi.component.send.mixin` for sending and implement the send function
+- `edi.component.check.mixin` for checking and implement the check function
+- `edi.component.validate.mixin` for validation and implement the validate function
+
+Also, the components must have the following elements that will be used to use the right component:
```
simahawk:
```suggestion
Also, the components may have the following elements that will be used to use the right component:
```

edi_component_oca/readme/CONFIGURE.md
```diff
@@ -0,0 +1,16 @@
...
+
+- `_backend_type`: code of the backend type
+- `_exchange_type`: code of the exchange type
+- `_usage`: Automatically set by the inheritade component
```
simahawk:
```suggestion
- `_usage`: Automatically set by the inherited component
```

edi_core_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -418,19 +422,35 @@ def _edi_send_via_email(
         composer.send_mail()
         return True
 
+    def _edi_config_field_relation(self):
```
simahawk, line 425:
```suggestion
    def _edi_execute_get_configs(self):
```

edi_core_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -418,19 +422,35 @@ def _edi_send_via_email(
...
+        # Link to edi.config fields to be used
+        # We can use a child fields like "self.partner_id.edi_config_ids"
+        return self.env["edi.configuration"]
+
+    def _execute_edi_config(self, trigger):
```
simahawk, line 430:
```suggestion
    def _edi_execute_config(self, trigger):
```

edi_core_oca/models/edi_exchange_type.py
```diff
@@ -88,6 +84,23 @@ class EDIExchangeType(models.Model):
         comodel_name="edi.exchange.type",
         compute="_compute_ack_for_type_ids",
     )
+    receive_model_id = fields.Many2one(
```
simahawk, line 87:
```suggestion
    receive_model_id = fields.Many2one(string="Receiver"
```
and so on for all the fields.

edi_core_oca/models/ir_model.py
simahawk: can you add this  "famous" comment? 😁

edi_core_oca/readme/ROADMAP.md
```diff
@@ -0,0 +1,7 @@
+## 14.0.1.0.0
+
+The module name has been changed from edi to edi_core_oca.
```
simahawk: not on 14.0 ;)

edi_core_oca/static/src/js/widget_edi.esm.js
```diff
@@ -1,5 +1,3 @@
-/** @odoo-module **/
```
simahawk, line 1: I cannot comment on the icon: still an outdated one ;)

edi_sale_oca/models/sale_order.py
```diff
@@ -12,6 +12,9 @@ class SaleOrder(models.Model):
         "edi.exchange.consumer.mixin",
     ]
 
+    def _edi_config_field_relation(self):
+        return self.partner_id.edi_sale_conf_ids
```
simahawk, line 16: hmm don't we risk to have 2 time the execution of the same conf? We are already using the listener component here. 🤔

simahawk (changes requested):

edi_sale_oca/models/sale_order.py
etobella: I don't see the problem. In which case a conf can be used twice?

_execute_edi_config, the function that handles this, is no longer defined in component or events :thinking:

edi_account_core_oca/readme/DESCRIPTION.md
etobella: Well, I thought that the icon was the old one and I made a new one, but after reviewing, I found that @LoisRForgeFlow already did a change on the logo. Here are both logos side by side. Right now, the new one is only applied on edi_core_oca (the old one is on the left):

Captura desde 2025-09-24 06-41-04

Shall I reuse the one made by @LoisRForgeFlow ?

etobella:
@simahawk One option would be to set a new edi abstract model for execution called `Component` and I can preset it everywhere. I will propose a different commit for it. WDYT? 

The same might happen for storage, endpoints, templates and so on. This way, this fields become "required".

edi_sale_oca/models/sale_order.py
simahawk: The consumer will make it run by default on write/create. At the same time the event handler will trigger the event. Am I missing anything?

edi_account_core_oca/readme/DESCRIPTION.md
simahawk: The 1st one on the left is the official one now on v18 and it looks really nice from my POV.
It comes from https://github.com/OCA/edi/blob/18.0/base_edi/static/description/icon.svg :wink:

etobella: I know, for that reason I named the author and my mistake :sob:  as I was used on the old icon on v17. I will use the one in base_edi :wink:

simahawk:
> @simahawk One option would be to set a new edi abstract model for execution called `Component` and I can preset it everywhere. I will propose a different commit for it. WDYT?
> 
> The same might happen for storage, endpoints, templates and so on. This way, this fields become "required".

Having one "Component Handler" to be assigned to all handler fields sounds good.

This handler will simply use the component machinery to look for the right thing to do.
IMO this simplifies everything:
* only one way to configure
* no need to override all the functions on the backend 

WDYT?

edi_sale_oca/models/sale_order.py
etobella: Yes, the event handler was removed :wink:

etobella:
:+1: I will take care of it (and prepare the right migration scripts too)

edi_sale_oca/models/sale_order.py
simahawk: isn't it still in edi_component_oca? :dizzy_face:

simahawk: aaaaah now I get it.... 🤣 
Since the listener is not there anymore the implementer can decide if to use the even or use the config api on the mixin. 👍

etobella:
@simahawk I did the same for edi_exchange_template_oca. I am not sure if that has sense in edi_storage_oca, as Storage depends a lot in the backend. not  in the exchange type.However, we can think about it in the future.

Also, I changed the exchange type view as it was too crowded
image

Also, the exchange_template field will appear only if the generator is exchange template, making it more comprehensive.

simahawk:
@etobella looks great! Thanks! I wanted to cleanup the view since quite some time :sweat_smile: 
As for latest changes it looks good. Nitpicking: some commits contain updates of the readme that will make harder to fwd/backport commits.

Regarding the storage: we'll have to do it to align the configuration.
Is not a blocker for this work tho. Maybe you can simply add a TODO in edi_storage?

etobella:
Well, storage is not merged in 18, so we shouldn't worry for now.

etobella:
Yes, I didn't notice the oca-gen-addon-readme skip was not there and I thought that the change were not imported. It should be fixed now.

edi_queue_oca/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1,2 @@
+- Firstname Lastname <email.address@example.org> (optional company website url)
```
simahawk: to be fixed

edi_queue_oca/__manifest__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2025 Dixmit
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+
+{
+    "name": "Edi Queue Oca",
+    "summary": """Set Queue Jobs on EDI""",
+    "version": "18.0.1.0.0",
+    "license": "LGPL-3",
+    "author": "Dixmit,Odoo Community Association (OCA)",
```
simahawk:
```suggestion
    "author": "Dixmit, Camptocamp, Odoo Community Association (OCA)",
```

edi_queue_oca/utils.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2020 ACSONE SA
```
simahawk, line 1:
```suggestion
# Copyright 2023 Camptocamp SA
```

edi_component_oca/models/edi_backend.py
```diff
@@ -0,0 +1,116 @@
+# Copyright 2025 Dixmit
```
simahawk, line 1:
```suggestion
# Copyright 2020 ACSONE SA
# Copyright 2021 Camptocamp
# Copyright 2025 Dixmit
```

edi_component_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2025 Dixmit
```
simahawk, line 2:
```suggestion
# Copyright 2025 Camptocamp SA
# Copyright 2025 Dixmit
```

edi_component_oca/models/edi_exchange_record.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2025 Dixmit
```
simahawk, line 2:
```suggestion
# Copyright 2025 Camptocamp SA
# Copyright 2025 Dixmit
```

edi_queue_oca/models/edi_exchange_record.py
```diff
@@ -0,0 +1,92 @@
+# Copyright 2025 Dixmit
```
simahawk, line 2:
```suggestion
# Copyright 2021 Camptocamp SA
# Copyright 2025 Dixmit
```

simahawk (changes requested):
Small remarks only 😉 
I have no time to test the last changes but I trust you did 😊 
Great work, thank you!

etobella:
Comments attended

etobella:
@acsonefho as you did a lot of similar work and part of it is inspired on it, we added you as co-author of the change.

I hope you agree with it.

simahawk (approved):

etobella:
@simahawk If you want, we can proceed with merge :wink:

simahawk:
GO GO GO :rocket: :rocket: :rocket:

etobella:
/ocabot merge patch

We will force the change of version Just in case

PR #145, opened 2025-03-04
[18.0][MIG] edi_stock_oca: Migration to 18.0

wouitmil:

PR #194, opened 2025-08-13
[18.0][MIG] edi_stock_oca

simahawk:
Replaces #145


henrybackman (approved):

edi_stock_oca/models/stock_picking.py
```diff
@@ -1,9 +1,41 @@
...
+            result = super()._action_done()
+        return result
+
+    def action_cancel(self):
+        with self._edi_exchange_event_trigger():
+            result = super().action_cancel()
+        return result
+
+    @contextmanager
+    def _edi_exchange_event_trigger(self):
```
grindtildeath, line 33: Using this, I'd assume you want "before state change" notification to happen even if the action returns an error and doesn't change the state?

FWIW We faced a similar situation in another project with a stored computed field where we did override `_write` to trigger the notification only if the field name was in the values dict.

grindtildeath (approved):
Looks ok :+1:

edi_stock_oca/models/stock_picking.py
simahawk: No, if it fails here the transaction fails and nothing happens.
As for the `_write` I thought of it but I decided to keep it simple.
If we have at least another case falling in this situation we could override `_write` in the consumer mixin directly and check if `state` is computed.

PR #204, opened 2025-09-23
[18.0] edi_oca: fix record form related_exchanges visibility

simahawk:
The page was not visible in any case. The condition was wrong. It should be hidden if no related exchange nor parent is set.


simahawk:
rebased

etobella (approved):

PR #206, opened 2025-09-30
[18.0][FIX] edi_core_oca: properly call model in _get_exec_handler

MiquelRForgeFlow:
....


MiquelRForgeFlow:
It's trivial, please.

PR #210, opened 2025-10-14
[18.0] edi_exchange_template: fix migration to 18.0.1.3.0

simahawk:
Fix misuse of the decorator. Without this the migration is completely broken.


simahawk:
@etobella this one slipped out of our sight :cry:

etobella (approved):

PR #212, opened 2025-10-14
[18.0] edi_endpoint: fix create_exchange_record helper

simahawk:
Setting the file content after creating the record might lead to errors when the exchange type is has quick_exec enabled. That's because the backend will try to process the file immediately on create but there's no file content yet.

Just populate the file content right away.


simahawk:
live and working smoothly since a while

/ocabot merge patch

PR #213, opened 2025-10-14
[18.0] edi_webservice_oca: fix component lookup permission

simahawk:
Avoid odoo.exceptions.AccessError when accessing webservice fields to match components.


etobella (approved):

simahawk:
added unit test

PR #214, opened 2025-10-14
[18.0] edi_core_oca: protect and track file content

simahawk:
So far the file could be edited in any state and w/o any trace of editing. This change brings:

* the file is frozen if it has been already processed or sent
* if you really need to change it you can force this by assigning the group ``group_edi_override_exchange_file_content`` to the user
* updates to the file are now tracked via checksum changes

edi-record-frozen-1

edi-record-frozen-2

NOTE: The flag is visible only in debug mode.

**TODO**
- [ ] fix tests
- [ ] avoid tracking of checksum on create


etobella (approved):

simahawk:
Finally green :green_apple: 

/ocabot merge patch

PR #216, opened 2025-10-28
[18.0][FIX] edi_core_oca: fix search logic

ThiagoMForgeFlow:
Maintaining this part of code can occur a visual issue. If the result variable has 80 records (default by Odoo) and the orig_ids is less than result, Odoo only shows the orig_ids, and we lost the previous work finding next records.

In a multicompany environment, if you only enable one company you won't see all the records that you should. Real example:

1. With only one company active you don't see all the records of that company.
screen1

2. You can only access all the records activiating all the companies.
screen2

Do the same logic proposed in 15: https://github.com/OCA/edi/pull/1147

@ForgeFlow

### DIFF
--- a/edi_core_oca/models/edi_exchange_record.py
+++ b/edi_core_oca/models/edi_exchange_record.py
@@ -616,8 +616,6 @@ class EDIExchangeRecord(models.Model):
             extend_ids = list(extend_query)
             result.extend(extend_ids[: limit - len(result)])
 
-        # Restore original ordering
-        result = [x for x in orig_ids if x in result]
         if set(orig_ids) != set(result):
             # Create a virgin query
             query = self.browse(result)._as_query()
--- a/edi_core_oca/tests/test_security.py
+++ b/edi_core_oca/tests/test_security.py
@@ -238,3 +238,76 @@ class TestEDIExchangeRecordSecurity(EDIBackendCommonTestCase):
         msg = rf"not allowed to access '{model._description}' \({model._name}\)"
         with self.assertRaisesRegex(AccessError, msg):
             child_exchange_record.with_user(self.user).read()
+
+    def test_search_pagination_with_inaccessible_middle_records(self):
+        """
+        Regression test:
+        If some records in the first page are filtered out due to access rules,
+        _search must fetch additional records from next pages without truncating them.
+        """
+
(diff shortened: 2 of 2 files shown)


DavidJForgeFlow (approved):

simahawk (changes requested):
Can we have a test ?
To ensure that:
1. final ordering is respected
2. company visibility fix works

Gracias! :)

etobella:
@ThiagoMForgeFlow gentle reminder

ThiagoMForgeFlow:
Hi @simahawk  and @etobella,

First of all, my apologies for the delayed response — I have been busy. I have just added a test that checks exactly what I mention in the PR. In this test, 100 `edi.exchange.record` records are created:

- The first 60 are linked to a res.partner record belonging to one company.
- The next 20 are linked to a res.partner record belonging to a new company.
- The last 20 are linked to a res.partner record belonging to the same company as the first one.

Then, when performing a search with a limit of 80 (simulating Odoo view pagination), with only the first company active and using a non-superuser, the expected result is the first 60 records and the last 20. Keeping the line of code proposed to be removed in this PR, the actual result is only the first 60 records, because the additional 20 records (from the second pagination) are not part of `orig_ids` and therefore are not included in the final result:
`result = [x for x in orig_ids if x in result]`

Ricardoalso (approved):

simahawk:
@ThiagoMForgeFlow LGTM. Just one "efficiency" point: do we really need to create 100 records? Assuming we can control the `limit`  I would say we can create way less records and save some computation time in the build. WDYT?

ThiagoMForgeFlow:
> @ThiagoMForgeFlow LGTM. Just one "efficiency" point: do we really need to create 100 records? Assuming we can control the `limit` I would say we can create way less records and save some computation time in the build. WDYT?

I’ve just applied the change so that only 3 exchange records are created and the limit is set to 2.

cyrilmanuel (approved):

ThiagoMForgeFlow:
Hi @simahawk could you do your final revision? Thanks :)

simahawk (approved):

LoisRForgeFlow:
@ThiagoMForgeFlow can you fwport this to 19?

PR #217, opened 2025-11-01
[18.0] edi_core_oca: fix access error on handler model

simahawk:
model.model few lines below will trigger an access error w/o this change.

TODO:
- [x] add tests 

@etobella am I wrong or we miss ACL completely for handler models? :thinking:


simahawk:
> @etobella am I wrong or we miss ACL completely for handler models? 🤔

I answer myself: they are abstract models, not need for that.
But we need that sudo call to access their `ir.model` properties.

simahawk:
Finally added tests :rocket:

PR #219, opened 2025-11-21
[18.0][FIX] edi_core_oca: move ordering inside conditional block

Ricardoalso:
Fixes major performance issues in edi_exchange_record_view_tree when grouping is enabled. The limit parameter only applies to the number of displayed group lines, but the grouping logic fetches all matching records, causing unnecessary reordering of the entire dataset when set(orig_ids) == set(result).

### DIFF
--- a/edi_core_oca/models/edi_exchange_record.py
+++ b/edi_core_oca/models/edi_exchange_record.py
@@ -616,9 +616,9 @@ class EDIExchangeRecord(models.Model):
             extend_ids = list(extend_query)
             result.extend(extend_ids[: limit - len(result)])
 
-        # Restore original ordering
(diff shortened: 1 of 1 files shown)


cyrilmanuel (approved):

santostelmo (approved):

simahawk (commented):
Seems legit. This goes in conflict w/ https://github.com/OCA/edi-framework/pull/216/
I've asked to add tests on that one. If that change is confirmed, this one can be dropped.

simahawk:
@Ricardoalso can you pls have a look at https://github.com/OCA/edi-framework/pull/216 ?

Ricardoalso:
Closed in favor of https://github.com/OCA/edi-framework/pull/216

PR #221, opened 2025-11-26
[16.0][FIX] edi_oca _cron_check_output_exchange_sync no check on quick_exec

santostelmo:
Records with type quick_exec=True will be processed immediately without waiting for the cron to pass by.

Therefore there's no point to lookup for pending output exchange record with quick_exec here: 
https://github.com/OCA/edi-framework/blob/33ae2216c8361999af5cf5d5f7865adebf135c20/edi_oca/models/edi_backend.py#L397

https://github.com/OCA/edi-framework/blob/33ae2216c8361999af5cf5d5f7865adebf135c20/edi_oca/models/edi_backend.py#L436


simahawk:
FTR here's a better fix https://github.com/OCA/edi-framework/pull/225

PR #222, opened 2025-12-06
[FIX] edi_storage_oca: Remove legacy storage component code

HviorForgeFlow:
Remove remaining code that was colliding when edi_component_oca module is installed in the system and not proper configuration is taken into account.


etobella (approved):

JordiMForgeFlow (approved):
we should merge this as soon as possible as it is causing errors

PR #225, opened 2026-01-20
[18.0] edi_core: clarify quick exec usage

simahawk:
Ignore quick_exec types when running automated actions via cron. We assume that quick exec records will be executed right away. You still have the possibility to enforce this using edi__force_generate ctx key.

At the same time, make the usage more clear in the UI: quick exec is available only if you want to generate the output automatically.


simahawk:
Besides... I wonder if we still need this auto-generate flag for the future.

etobella (approved):
TEchnically looks ok

PR #226, opened 2026-01-26
[18.0][FIX] edi_storage_oca: add event listener

ArnauCForgeFlow:
During the migration of edi_storage_oca to the new EDI version, the listener that executes once the EDI record has been processed was omitted.


LoisRForgeFlow (changes requested):
Can you add a test case to avoid regressions again?

edi_storage_oca/__manifest__.py
```diff
@@ -12,7 +12,7 @@
     "license": "LGPL-3",
     "website": "https://github.com/OCA/edi-framework",
     "author": "ACSONE,Odoo Community Association (OCA)",
-    "depends": ["edi_core_oca", "fs_storage"],
+    "depends": ["edi_core_oca", "fs_storage", "edi_component_oca"],
```
LoisRForgeFlow, line 15: @etobella Could we ask for your opinion here? there was a regression, but to revert it we need a new dependency... not sure if you would prefer this in a glue module. Thoughts?

ArnauCForgeFlow:
@LoisRForgeFlow tests added

etobella (changes requested):
If you need a component, you should add a new module (glue module)

etobella:
Silly question, can you get the same result without components?

ArnauCForgeFlow:
I am using this module together with document_quick_access_folder_auto_classification, which is responsible for processing the file. I added the listener to the edi_storage_oca module because I understand it to be a generic behavior that should be used for all records that use the edi_storage_oca backend.

It works because the document_quick_access_folder_auto_classification module adds edi_component_oca, so I assume that the listener does not work on its own unless the component dependency is added.

etobella:
Yes, but that's because I did not had the time to remove the dependency :wink:

ArnauCForgeFlow:
So, your suggestion is create a glue module with the listenner and the component dependency?

etobella:
my suggestion would be:

1- Can we do it without components? If yes, do it
2- If no, make a glue module

simahawk:
Short term, I would:

* move `_trigger_edi_event_make_name` from the component module to core
* hook to `_trigger_edi_event` to do what you need

Long term, this makes me think that we should probably have an explicit handler for state transition on the type.
Something like `state_change_handler_model_id`. With that we can:

* move `_trigger_edi_event_make_name` from the component module to core
* modify `_trigger_edi_event` to do look for this handler and make it call the right `on_*` method


Opinions?

ArnauCForgeFlow:
Closing this PR, since this is now replaced by: https://github.com/OCA/edi-framework/pull/280

PR #227, opened 2026-01-28
[ADD] edi_mail_import_oca

etobella:
Usually, when we think about EDI entrances, we think about SFTP or API, however, there is an standard way to receive files that is emails.

This module tries to offer a way to reuse EDI and mail interface in order to handle everything.

By default it offers 2 ways to receive files, to get the file directly from the received or just  pass the full email as JSON (might be needed if there are multiple files)

@simahawk WDYT?


simahawk:
Just for reference, the feedback is here https://github.com/OCA/edi-framework/pull/229 :)

PR #215, opened 2025-10-15
[19.0][MIG] edi_core_oca

etobella:
Done the history rewritting for not losing the commits from edi_oca and map them in edi_core_oca (Since the split, obviously).

I added JS tests :wink: 

I avoided the use of odoo_test_helper

PR #229, opened 2026-02-02
[19.0][MIG] edi_mail_import_oca

etobella:
Depends on https://github.com/OCA/edi-framework/pull/215


edi_mail_import_oca/models/edi_exchange_record.py
```diff
@@ -0,0 +1,67 @@
...
+
+    @api.model
+    def message_new(self, msg_dict, custom_values=None):
+        record = super().message_new(
+            msg_dict,
+            custom_values=custom_values,
+        )
+        if record.type_id.direction != "input":
+            raise UserError(
+                self.env._("Received email for non-incoming exchange type.")
```
simahawk: wouldn't be better to add a validator for this? If you set an alias on the type it must be of type input. If not, you will never reach this point, no?

edi_mail_import_oca/models/edi_exchange_type.py
```diff
@@ -0,0 +1,27 @@
...
+    _inherit = ["edi.exchange.type", "mail.alias.mixin"]
+
+    mail_as_attachment = fields.Boolean(
+        string="Import Email as an Attachment",
+    )
+
+    def _alias_get_creation_values(self):
+        values = super()._alias_get_creation_values()
+        values["alias_model_id"] = (
+            self.env["ir.model"].sudo()._get("edi.exchange.record").id
```
simahawk: sudo is not necessary here. Besides if you need only the id just use `_get_id`

edi_mail_import_oca/readme/DESCRIPTION.md
```diff
@@ -0,0 +1 @@
+This module allows to process mails received as edi.exchange.records to process them.
```
simahawk, line 1: maybe add a configuration file?

edi_mail_import_oca/models/edi_exchange_record.py
```diff
@@ -0,0 +1,67 @@
...
+        else:
+            content = False
+            filename = False
+            for attachment in msg_dict.get("attachments", []):
+                if re.match(
+                    record.type_id.exchange_filename_pattern or ".*",
+                    attachment.fname,
+                    re.IGNORECASE,
+                ):
+                    content = self._process_email_attachment(attachment)
```
simahawk: I cannot find this method in Odoo 🤔

edi_mail_import_oca/models/edi_exchange_record.py
```diff
@@ -0,0 +1,67 @@
...
+                    }
+                )
+            record._set_file_content(json.dumps(new_message_dict))
+            record.edi_exchange_state = "input_received"
+        else:
+            content = False
+            filename = False
+            for attachment in msg_dict.get("attachments", []):
+                if re.match(
+                    record.type_id.exchange_filename_pattern or ".*",
```
simahawk: so the deal is that one of the attachments of the email should match the pattern, right?

edi_mail_import_oca/views/edi_exchange_type.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2026 Dixmit
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record model="ir.ui.view" id="edi_exchange_type_view_form">
+        <field name="model">edi.exchange.type</field>
+        <field name="inherit_id" ref="edi_core_oca.edi_exchange_type_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//group[@name='config']" position="after">
```
simahawk: maybe better to move this to its own page?
IMO it should mention in an info box that the file pattern will be used to match attachments in the emails.

edi_mail_import_oca/models/edi_exchange_type.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2026 Dixmit
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+import ast
+
+from odoo import fields, models
+
+
+class EdiExchangeType(models.Model):
+    _name = "edi.exchange.type"
+    _inherit = ["edi.exchange.type", "mail.alias.mixin"]
```
simahawk: Thinking out loud: instead of polluting the exchange type model, wouldn't be interesting to have a specific record for this? Eg: "EDI mail gateway" or something similar.
Easier to maintain and to configure for the end user probably. WDYT?

simahawk (changes requested):
Interesting approach! Thanks!

edi_mail_import_oca/models/edi_exchange_record.py
etobella: yes

edi_mail_import_oca/models/edi_exchange_record.py
etobella: https://github.com/OCA/edi-framework/pull/229/changes#diff-f220327b1b0df1fb3d54b634ab9871d1f771b9b7c8df498039350bd155fb83f2R55

edi_mail_import_oca/models/edi_exchange_type.py
etobella: Or maybe using `edi.backend`. At the end, that is the goal of backends, isn't it?

edi_mail_import_oca/readme/DESCRIPTION.md
etobella: Sure!

etobella:
@simahawk After your comments, I do agree that the best approach would be to set everything on the backend and do the same thing we do with Storage.

Also, I improved REadmes and information inside the backend to make it easier to handle (actually, it would be nice to include similar information in other modules to make it easier for users. Thanks for your comments, IMO they improved the solution.

image

edi_mail_import_oca/models/edi_exchange_record.py
simahawk: ahah, I don't why I was looking for this method in Odoo.... I assumed it was coming from the mixin probably :sweat_smile:

edi_mail_import_oca/views/edi_backend.xml
```diff
@@ -0,0 +1,67 @@
...
+                    <group>
+                        <group>
+                            <div name="alias_def" colspan="2" class="pb-2">
+                                <!-- Always display the whole alias in edit mode. It depends in read only -->
+                                <!-- Need to add alias_id in view for getting alias_domain_id by default -->
+                                <field name="alias_id" invisible="1" />
+                                <label
+                                    for="alias_name"
+                                    class="fw-bold o_form_label"
+                                    string="Create tasks by sending an email to"
```
simahawk:
```suggestion
                                    string="Create EDI records by sending an email to"
```

edi_mail_import_oca/models/edi_backend.py
```diff
@@ -0,0 +1,35 @@
...
+        values["alias_model_id"] = self.env["ir.model"]._get("edi.exchange.record").id
+        if self.id:
+            values["alias_defaults"] = defaults = ast.literal_eval(
+                self.alias_defaults or "{}"
+            )
+            defaults["backend_id"] = self.id
+        return values
+
+    def _mail_exchange_type_pending_input_domain(self):
+        """Domain for retrieving input exchange types for emails."""
```
simahawk, line 23: I would add some more conventional constraint here. 
What if we add specific type of backend `Email` and we rely on the fact that exc types refer to that type.
I would make it more clear from the configuration point of view.

edi_mail_import_oca/models/edi_exchange_record.py
```diff
@@ -0,0 +1,90 @@
...
+        backend_id = custom_values.get("backend_id") if custom_values else None
+        if backend_id:
+            backend = self.env["edi.backend"].browse(backend_id)
+            content = False
+            filename = False
+            if (
+                not custom_values.get("type_id")
+                and not backend.full_mail_exchange_type_id
+            ):
+                types = self.env["edi.exchange.type"].search(
```
simahawk: if we rely on the fact that we can have multiple files in an email and multiple types configured, we should support multiple creation of records.

Hence, here we should collect all new record values and call super on each set of values at the end.
Also, each record should be related to a single parent record (conventionally I would use the 1st one).

edi_mail_import_oca/models/edi_exchange_record.py
```diff
@@ -0,0 +1,90 @@
...
+                        ):
+                            content = self._process_email_attachment(attachment)
+                            filename = attachment.fname
+                            custom_values["type_id"] = exchange_type.id
+                            break
+                    if content:
+                        custom_values["type_id"] = exchange_type.id
+                        break
+            else:
+                if backend.full_mail_exchange_type_id:
```
simahawk: If I get it right you want the possibility to just store the full content of the email. Correct?

If that's the case, I would move this configuration on the type itself.
Something like:

```
mail_record_policy = fields.Selection(
    [
        ("full", "Only archive the full mail message"),
        ("pattern", "Match single filename pattern"),
    ]
)
```
This field will be visible only for the backend type `email` and will command how this record creation will work.

As for the full json storage: IMO attachments should not be incapsulated into the json data.
We should store them as attachments of the exchange record.

WDYT?

edi_mail_import_oca/models/edi_exchange_record.py
```diff
@@ -0,0 +1,90 @@
...
+                        {
+                            "info": attachment.info,
+                            "data": self._process_email_attachment(attachment),
+                            "fname": attachment.fname,
+                        }
+                    )
+                content = json.dumps(new_message_dict)
+                filename = "email_message.json"
+        if not custom_values or "type_id" not in custom_values:
+            _logger.warning(
```
simahawk: is there a need for this logging if you raise an error?

edi_mail_import_oca/models/edi_exchange_record.py
etobella: :+1:

edi_mail_import_oca/models/edi_exchange_record.py
etobella: Well, I understand the idea, however, we cannot do that, as the message_new should return just one record and is attaching some extra information... however, I will find a solution for that. don't worry :wink:

etobella:
@simahawk can you make a small review so we can proceed with merge?

simahawk (commented):
Last nitpicking remarks ;)

SilvioC2C (approved):

PR #231, opened 2026-02-09
[16.0] [FIX] edi_stock_oca: fix view

luisDIXMIT:
With this change, we fix the input field being too small.
image

@etobella


etobella (approved):
Thanks! IT is a small fix, but quite useful.

etobella:
/ocabot merge patch

This is a small view fix.

Can you review in 17 and 18? :pray:

PR #228, opened 2026-01-30
[18.0][ADD] edi_storage_component_oca

ArnauCForgeFlow:
Glue module between edi_storage_oca and edi_component_oca.

It implements the EDI processing listener using the OCA component framework, allowing custom behaviors to be executed when EDI records are processed.

PR #232, opened 2026-02-12
[18.0][IMP] edi_core_oca: add listener action

ArnauCForgeFlow:
Move **_trigger_edi_event_make_name** method to edi_core_oca module in order to make the listener usable without requiring the edi_component_oca module


edi_core_oca/models/edi_exchange_record.py
```diff
@@ -688,3 +683,20 @@ def write(self, vals):
     def action_exchange_generate_send_chained(self):
         self.action_exchange_generate()
         self.action_exchange_send()
+
+    def _trigger_edi_event_make_name(self, name, suffix=None):
```
simahawk: should be removed from the component module

edi_core_oca/models/edi_exchange_record.py
```diff
@@ -688,3 +683,20 @@ def write(self, vals):
...
+            name=name,
+            suffix=("_" + suffix) if suffix else "",
+        )
+
+    def _trigger_edi_event(self, name, suffix=None, target=None, **kw):
+        method_name = self._trigger_edi_event_make_name(name, suffix)
+        target = target or self
+
+        if hasattr(target, method_name):
+            target.with_delay().action_exchange_listener(method_name, **kw)
```
simahawk: we don't want to use jobs here

edi_core_oca/models/edi_exchange_record.py
```diff
@@ -688,3 +683,20 @@ def write(self, vals):
...
+
+    def action_exchange_listener(self, method_name, **kw):
+        if hasattr(self, method_name):
```
simahawk: hmm AFAIS here we would demand integrators to add hooks to the exc record. I don't think that's desirable.

Can't we leverage EDI configuration for this? The machinery on the mixin is already there.
Otherwise, as proposed in the other PR, we should be explicit and declare handlers on the type.

edi_queue_oca/data/job_function.xml
```diff
@@ -19,6 +19,11 @@
         <field name="method">action_exchange_process</field>
         <field name="channel_id" ref="channel_edi_exchange" />
     </record>
+    <record id="job_fun_exchange_record_listener" model="queue.job.function">
```
simahawk: I don't think we need this. It's up to the implementer to use jobs or not.

simahawk (changes requested):

ArnauCForgeFlow:
@simahawk Thanks for the review!

So, basically, your suggestion is moving the helper method _trigger_edi_event_make_name from the component module to the core, and then simply calling the _trigger_edi_event hook from the other modules?

simahawk:
> @simahawk Thanks for the review!
> 
> So, basically, your suggestion is moving the helper method _trigger_edi_event_make_name from the component module to the core, and then simply calling the _trigger_edi_event hook from the other modules?

yes. And we should move the hook to the consumer mixin so that the logic won't be put into generic models like exc record.

ArnauCForgeFlow:
> > @simahawk Thanks for the review!
> > So, basically, your suggestion is moving the helper method _trigger_edi_event_make_name from the component module to the core, and then simply calling the _trigger_edi_event hook from the other modules?
> 
> yes. And we should move the hook to the consumer mixin so that the logic won't be put into generic models like exc record.

The problem with moving the hook to the consumer mixin is that for input exchanges that fail early, there is no record_id yet. For example, in edi_storage_oca, even if the process fails, we still need to execute the hook to move the file from 'pending' directory to 'error' directory. Any idea on how to handle this?

Thanks!

simahawk:
@ArnauCForgeFlow good point. Well, that's why components have been used so far :smile: 

A possible approach is the one I took on endpoint_route_handler to check cross models that we wouldn't have any routing conflict. See https://github.com/OCA/web-api/blob/18.0/endpoint_route_handler/models/endpoint_route_handler.py#L90

We could collect all models inheriting from consumer mixin and trigger their handlers if any.
Something like that:

```
class EDIExchangeConsumerMixin(models.AbstractModel):
    _name = "edi.exchange.consumer.mixin"
    [...]

    @api.model
    def _get_edi_consumer_models(self):
        global EDI_CONSUMER_MODELS
        if EDI_CONSUMER_MODELS.get(self.env.cr.dbname):
            return EDI_CONSUMER_MODELS.get(self.env.cr.dbname)
        models = []
        for model in self.env.values():
            if (
                model._name != self._name
                and not model._abstract
                and self._name in model._inherit
            ):
                models.append(model._name)
        EDI_CONSUMER_MODELS[self.env.cr.dbname] = models
        return models

    def edi_trigger_consumer_models(self, trigger, records=None):
        """Trigger EDI config execution on all related models."""
        for model_name in self._get_edi_consumer_models():
            model = self.env[model_name]
            if hasattr(model, trigger):
                getattr(model, trigger)(records)
```
WDYT?

simahawk:
Answering myself: this can lead to overlapping handlers as you could have the same hook for different exchanges.

Could we leverage `edi.configuration` then?
By default edi config are meant to work w/ the partners linked to a record/model and a trigger.
We could have a generic configuration that does not need a partner but only a trigger and a model.
This will leave the door opened for manual configuration and to place the final handler where you prefer.

ArnauCForgeFlow:
Hi @simahawk Thanks for the suggestion! I've been looking into the edi.configuration approach, but to be honest, I'm a bit lost, so I'd love a bit more context on the implementation details.

- Which method should be the trigger? on_edi_exchange_done / on_edi_exchange_error, or _trigger_edi_event?
- In which model should I place the hooks? A custom one, edi.exchange.record, or edi.exchange.consumer.mixin?

Any quick example or guidance would be appreciated. Thanks!

simahawk:
All starts with [notify_action_complete](https://github.com/OCA/edi-framework/blob/18.0/edi_core_oca/models/edi_exchange_record.py#L509C9-L509C31). This must be called all the times an "official" action is done.
In turns, it calls `_trigger_edi_event`, which should take care of what we are discussing now.

There we should trigger the event an all related records and all consumer models.
The trigger on related records partly implemented  is since the split https://github.com/OCA/edi-framework/blob/18.0/edi_core_oca/models/edi_exchange_record.py#L525 but it's incomplete as `_trigger_edi_event` does nothing ATM.

Hence, for related records, we should do what is done already for state change, create and write https://github.com/OCA/edi-framework/blob/18.0/edi_core_oca/models/edi_exchange_consumer_mixin.py#L449
There's it's easier because the lookup of edi.config is done already via `_edi_config_field_relation` (eg: https://github.com/OCA/edi-framework/blob/b7ca2bc3e2744b1fe30ba811bfd15d609d748fe8/edi_sale_oca/models/sale_order.py#L16). 

When we don't have a record, we have to decide how to lookup for edi.confg. Today we always rely on a m2m relation via partner (we have a m2m relation to be able to reuse configs). We should probably state which configs are meant to run globally for a given model (eg: w/ a flag `global` or `non_record_related`, whatever).

I hope it's clear. Otherwise we can organize a call :wink:

edi_core_oca/models/edi_configuration.py
```diff
@@ -67,6 +67,12 @@ class EdiConfiguration(models.Model):
         help="""Used to do something specific here.
         Receives: operation, edi_action, vals, old_vals.""",
     )
+    is_global = fields.Boolean(
+        string="Global Configuration",
+        help="If checked, this configuration will be executed for all records, "
+        "regardless of the partner.",
```
simahawk, line 73:
```suggestion
        "regardless of any related odoo record.",
```

edi_component_oca/models/edi_exchange_record.py
```diff
@@ -8,15 +8,9 @@
...
-    def _trigger_edi_event_make_name(self, name, suffix=None):
-        return "on_edi_exchange_{name}{suffix}".format(
-            name=name,
-            suffix=("_" + suffix) if suffix else "",
-        )
-
     def _trigger_edi_event(self, name, suffix=None, target=None, **kw):
         """Trigger a component event linked to this backend and edi exchange."""
-        name = self._trigger_edi_event_make_name(name, suffix=suffix)
+        method_name = self._trigger_edi_event_make_name(name, suffix=suffix)
```
simahawk, line 13:
```suggestion
        event_name = self._trigger_edi_event_make_name(name, suffix=suffix)
```
to be compliant w/ your change below ;)

edi_storage_oca/data/edi_configuration.xml
```diff
@@ -0,0 +1,24 @@
+<odoo noupdate="1">
+    <record id="edi_config_trigger_storage_done" model="edi.configuration.trigger">
+        <field name="name">On record exchange done</field>
```
simahawk, line 3: aren't these triggers generic enough to stay in the core module? 🤔

ArnauCForgeFlow:
Hi, @simahawk I've applied the suggested changes. Please let me know if this aligns with what you had in mind. I’ve also included in an extra commit how edi_storage_oca will look with this change to show the full integration (if we merge, this commit will be deleted and added to the other PR). 

Please let me know if I missed anything or misunderstood something. Thanks!

edi_storage_oca/models/edi_exchange_record.py
```diff
@@ -13,3 +19,67 @@ class EDIExchangeRecord(models.Model):
         string="FS Storage",
         help="Record created from a file found in this FS storage",
     )
+
+    def _move_file(self, storage, from_dir_str, to_dir_str, filename):
```
simahawk, line 23: I would move this to fs_storage as we are dragging this from project to project, module to module

edi_storage_oca/models/edi_exchange_record.py
```diff
@@ -13,3 +19,67 @@ class EDIExchangeRecord(models.Model):
...
+
+    def _add_post_commit_hook(
+        self, move_func, storage, sftp_filepath, sftp_destination_path
+    ):
+        """Add hook after commit to move the file when transaction is over."""
+        self.env.cr.postcommit.add(
+            functools.partial(move_func, storage, sftp_filepath, sftp_destination_path)
+        )
+
+    def on_edi_exchange_done(self):
```
simahawk, line 48:
```suggestion
    def _storage_on_edi_exchange_done(self):
```

edi_storage_oca/models/edi_exchange_record.py
```diff
@@ -13,3 +19,67 @@ class EDIExchangeRecord(models.Model):
...
+                return res
+            res = self._move_file(storage, pending_dir, done_dir, file)
+            if not res:
+                # If a file previously failed it should have been previously
+                # moved to the error dir, therefore it is not present in the
+                # pending dir and we need to retry from error dir.
+                res = self._move_file(storage, error_dir, done_dir, file)
+        return res
+
+    def on_edi_exchange_error(self):
```
simahawk, line 72:
```suggestion
    def _storage_on_edi_exchange_error(self):
```
I'm not a fan of adding this custom logic to the record, having it on the component was better. 
However, TBH, I have no better place to suggest 🤷

ArnauCForgeFlow: Yes, I agree with you, I think it's much cleaner to keep using components and move that logic into a glue module, as I did in my first approach.

simahawk: we can, but at least we should prefix w/ proper naming (eg: `_storage_*`).

@etobella your POV?

ArnauCForgeFlow: @etobella friendly reminder!

etobella: If we need to add a component, I prefer a glue module. I know it is not the best solution, but it is better to avoid dependancies that are not needed. About the naming, edi_storage_event or something like that looks fine for me.

simahawk:
@ArnauCForgeFlow FYI I'm getting back to this finally :hourglass_flowing_sand:  
If you don't mind I'll take over this change preserving parts of what you are done. We are not far! :love_you_gesture: 
Thanks for you work and for trying to read my imagination :laughing:

simahawk:
@ArnauCForgeFlow there you go https://github.com/OCA/edi-framework/pull/275.
I've left our the storage part for now.

simahawk:
For the storage this is what I would do:

* move `_move_file` to `fs_storage.models.fs_storage` where `self` is the storage itself
* move `on_edi_exchange_done` and `on_edi_exchange_done` to `edi_storage.models.edi_backend` and rename them as `_storage_on_*`
* add default inactive global edi.configuration that call those methods in the right way. The snippet should mention in a comment that before enabling it a backend must be chosen.

Then you can drop #228 :wink: 

WDYT?

ArnauCForgeFlow:
HI @simahawk The changes look good, thanks! I'll adapt my work on edi_storage_oca and open a new PR that depends on https://github.com/OCA/edi-framework/pull/275, following the refactoring you suggested. I'll get to it as soon as possible

PR #233, opened 2026-02-13
[18.0][IMP] edi_core_oca:  add backend-controlled auto-cleanup for exchange records

Ricardoalso:
This pull request introduces automated maintenance for EDI exchange records. The most significant changes are the addition of scheduled jobs to archive and delete old EDI records and the introduction of an `active` flag for EDI records.

Changes:
- Add "Auto Cleanup" tab in backend form view for easy configuration
- Add archive cron job that respect backend-specific retention policies
- Add delete cron job that respect backend-specific retention policies
- Add "Archived" filter to exchange record search view

Benefits:
- Each backend can have different retention policies
- Setting fields to 0 (default value) disables the respective cleanup behavior
- More flexible data retention management
- Better control over storage and performance optimization

The cron jobs iterate through backends with configured retention
settings, applying appropriate cutoff values.


etobella (changes requested):
If we want to do that, we need to at least add a way to decide which records can be archived. IMO, we should add a flag on exchange type or backends for that.

Ricardoalso:
> If we want to do that, we need to at least add a way to decide which records can be archived. IMO, we should add a flag on exchange type or backends for that.

Hey @etobella, thanks for the early review, you were too fast for me! 😄

Following @simahawk’s suggestion, I added the configuration to the edi.backend model and kept them disabled by default.

simahawk:
> If we want to do that, we need to at least add a way to decide which records can be archived. IMO, we should add a flag on exchange type or backends for that.

You can still edit the cron to search for records by specific type.
We could add a domain to the backend but for the moment I would keep it simple.

edi_core_oca/views/edi_backend_views.xml
```diff
@@ -57,7 +57,14 @@
                         <field name="company_id" groups="base.group_multi_company" />
                     </group>
                     <!-- Hook to add more config -->
-                    <notebook />
+                    <notebook>
+                        <page name="auto_cleanup" string="Auto Cleanup">
```
simahawk:
```suggestion
                        <page name="auto_cleanup" string="Records retention">
```

simahawk (approved):

cyrilmanuel (approved):

edi_core_oca/models/edi_backend.py
```diff
@@ -62,6 +62,18 @@ class EDIBackend(models.Model):
     )
     active = fields.Boolean(default=True)
     company_id = fields.Many2one("res.company", string="Company")
+    auto_archive_records_after_days = fields.Integer(
+        string="Auto-archive records after (days)",
+        default=0,
+        help="Automatically archive EDI exchange records after X days. "
+        "Set to 0 to disable auto-archiving.",
```
SilvioC2C: ### SOLVED :white_check_mark: 

For consistency with the cron's code:
```suggestion
        "Set to <= 0 to disable auto-archiving.",
```

edi_core_oca/models/edi_backend.py
```diff
@@ -62,6 +62,18 @@ class EDIBackend(models.Model):
...
+    )
+    auto_delete_records_after_days = fields.Integer(
+        string="Auto-delete archived records after (days)",
+        default=0,
+        help="Automatically delete archived EDI exchange records after X days. "
+        "Set to 0 to disable auto-deletion.",
```
SilvioC2C: ### SOLVED :white_check_mark: 

For consistency with the cron's code:
```suggestion
        "Set to <= 0 to disable auto-deletion.",
```

edi_core_oca/views/edi_exchange_record_views.xml
```diff
@@ -265,6 +265,12 @@
                     help="Show all records created in the last 7 days"
                 />
                 <separator />
+                <filter
+                    string="Archived"
+                    name="filter_archived"
+                    domain="[('active', '=', False)]"
+                />
```
SilvioC2C, line 277: ### SOLVED :white_check_mark: 

It could be useful to include filters for both active and archived records:
```suggestion
                <filter
                    string="Active"
                    name="filter_active"
                    domain="[('active', '=', True)]"
                />
                <filter
                    string="Archived"
                    name="filter_archived"
                    domain="[('active', '!=', True)]"
                />
```

SilvioC2C (approved):

simahawk:
@etobella ping :)

etobella (approved):
Technical review

Thanks!

PR #237, opened 2026-02-20
[18.0][FIX] edi_core_oca: notify action completion based on exchange state

Ricardoalso:
CI failing because of FakeModelLoader; addressing it in  https://github.com/OCA/edi-framework/pull/238

**EDIT**
Cherry-picked the FakeModelLoader commit in here in order to let the CI run, but I will rebase it as soon the FakeModelLoader PR gets merged  https://github.com/OCA/edi-framework/pull/238

### DIFF
--- a/edi_core_oca/models/edi_backend.py
+++ b/edi_core_oca/models/edi_backend.py
@@ -142,7 +142,8 @@ class EDIBackend(models.Model):
                         "exchange_error_traceback": traceback,
                     }
                 )
-        exchange_record.notify_action_complete("generate", message=message)
+        if exchange_record.edi_exchange_state == "output_pending":
+            exchange_record.notify_action_complete("generate", message=message)
         return message
 
     # TODO: unify to all other checkes that return something
@@ -258,7 +259,11 @@ class EDIBackend(models.Model):
                     "exchanged_on": fields.Datetime.now(),
                 }
             )
-        exchange_record.notify_action_complete("send", message=message)
+        if exchange_record.edi_exchange_state in [
+            "output_sent",
+            "output_sent_and_processed",
+        ]:
+            exchange_record.notify_action_complete("send", message=message)
         return res
 
     def _swallable_exceptions(self):
@@ -466,7 +471,8 @@ class EDIBackend(models.Model):
                 exchange_record._notify_error("process_ko")
             elif state == "input_processed":
                 exchange_record._notify_done()
-        exchange_record.notify_action_complete("process", message=message)
+        if exchange_record.edi_exchange_state == "input_processed":
+            exchange_record.notify_action_complete("process", message=message)
         return res
 
     def _exchange_process(self, exchange_record):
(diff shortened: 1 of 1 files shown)


etobella (approved):
It makes sense

guewen (approved):

gurneyalex (approved):

simahawk:
@Ricardoalso pls rebase

edi_core_oca/models/edi_backend.py
```diff
@@ -142,7 +142,8 @@ def exchange_generate(self, exchange_record, store=True, force=False, **kw):
                         "exchange_error_traceback": traceback,
                     }
                 )
-        exchange_record.notify_action_complete("generate", message=message)
+        if exchange_record.edi_exchange_state == "output_pending":
```
simahawk, line 145: I don't think this is correct. For instance, in this case, there won't be any notification for `validate_ko`.
A long time ago I refactored this part to always notify w/ `notify_action_complete` no matter what.
It doesn't not matter _how_ it completes, it must always notify to trigger events or post a message somewhere.
What are you trying to solve in particular?

simahawk (changes requested):

edi_core_oca/models/edi_backend.py
Ricardoalso: For line 145, I agree with your feedback; my change is incorrect in restricting notification.

his PR targets the same underlying issue referenced in [PR #240](https://github.com/OCA/edi-framework/pull/240/changes). The goal was to avoid calling notify_action_complete on an exchange_record with a dead cursor. I split the PRs to keep things explicit, but introducing a raise for OperationalError or IntegrityError ensures that if a cursor dies, the function flow is interrupted, and neither notification nor downstream hooks are triggered.

Without exception handling, the flow would continue, notifications would be sent, and hooks would run—which is exactly what we wanted to prevent when encountering these database errors. If you see other scenarios or states where this might not be sufficient, I’m open to suggestions!

simahawk: but if the errors are raised then you don't need this, am I wrong?

Ricardoalso: If they are raised, there’s indeed no need, as `notify_action_complete` would not be reached.

cyrilmanuel (approved):

edi_core_oca/models/edi_backend.py
```diff
@@ -258,7 +259,11 @@ def exchange_send(self, exchange_record):
                     "exchanged_on": fields.Datetime.now(),
                 }
             )
-        exchange_record.notify_action_complete("send", message=message)
+        if exchange_record.edi_exchange_state in [
```
simahawk, line 262: @Ricardoalso based on your feedback above, if we merge #240 there's no reason to check the state here. Unless, we expect that we can still reach this line under some other circumstance.
If we expect that we need a comment to explain why.
And if we reach this state and we don't trigger the event it means something is fishy and I would at least return a proper message at the end so that at least ad job level we have a trace.
As an alternative we raise a UserError and let the job fail.
WDYT?

Ricardoalso: I agree. As https://github.com/OCA/edi-framework/pull/240 has been merged, these state checks are unnecessary and may hide unexpected behavior by dropping the completion event (notify_action_complete).

This was some time ago, but I do not recall any specific cases apart from the dead cursor issue, which has already been addressed by https://github.com/OCA/edi-framework/pull/240. I suggest we close this PR.

PR #239, opened 2026-02-23
[19.0][MIG] edi_account_core_oca

JasminSForgeFlow:
Standard Migration

@ForgeFlow


Reyes4711-S73 (approved):
LGTM

Segui-S73 (approved):
LGTM

JordiMForgeFlow (approved):

JordiMForgeFlow:
@simahawk @etobella could this one get merged? thanks!

edi_account_core_oca/hooks.py
```diff
@@ -0,0 +1,10 @@
+def pre_init_hook(env):
```
etobella, line 1: Not required anymore.

etobella (commented):
@JordiMForgeFlow as we are on a new migration, I think we should rename this module as edi_account_oca, the original one is not required anymore.

Can you handle the merge on openupgrade?

JordiMForgeFlow:
@etobella then I assume we also have to merge the listeners added in https://github.com/OCA/edi-framework/pull/265 to this one right?

@JasminSForgeFlow could you attend the changes when possible?

etobella:
Wait! I assumed that the other one had no code inside, if it has, we should not merge them, as this module should never require listeners, components or queue. maybe the right new should be "edi_account_listener_oca" and "edi_account_oca", however, I think it doesn't make sense to do it now.

JordiMForgeFlow:
Exactly, the [edi_account_oca](https://github.com/OCA/edi-framework/pull/265) is depending on component to add the listeners, just like edi_stock_oca or edi_purchase_oca. For me it is fine to keep it as it is then.

etobella:
For me the names could be improved to avoid mistakes....

JordiMForgeFlow:
@etobella so just to clarify, you would rename this one to "edi_account_oca" and https://github.com/OCA/edi-framework/pull/265 to "edi_account_listener_oca"? Then we should also align the other modules to use the same structure (edi_stock_oca, edi_purchase_oca,...)

simahawk:
Hmm do we really need 2 modules? 
It's probably time to align over a call on what we want to do once for all.
IMO we should finish https://github.com/OCA/edi-framework/pull/232 first and agree on a common way to handle events w/o components otherwise we will always have an half backed solution in the core and 2 ways to things everywhere.

DavidJForgeFlow (approved):

MiquelRForgeFlow (approved):
@JasminSForgeFlow could you rebase?

JasminSForgeFlow:
> @JasminSForgeFlow could you rebase?

Done, Thanks

simahawk (changes requested):
> Hmm do we really need 2 modules? It's probably time to align over a call on what we want to do once for all. IMO we should finish #232 first and agree on a common way to handle events w/o components otherwise we will always have an half backed solution in the core and 2 ways to things everywhere.

Still waiting for an answer ;) 

Good occasion to discuss: OCA days ;)

PR #238, opened 2026-02-20
[18.0][FIX] edi_*_oca: New way for FakeModelLoader

Ricardoalso:
Followup of https://github.com/OCA/odoo-test-helper/pull/39

PR #240, opened 2026-02-23
[18.0][FIX] edi_core_oca: handle OperationalError and IntegrityError exceptions to prevent finally …

Ricardoalso:
…block execution in aborted transaction state

In failed transactions, the cursor enters an aborted state where any further SQL (including finally block writes) fails. Previously unhandled, causing cascading errors.

> Cherry-picked the FakeModelLoader commit in here in order to let the CI run, but I will rebase it as soon the FakeModelLoader PR gets merged https://github.com/OCA/edi-framework/pull/238

### DIFF
--- a/edi_core_oca/models/edi_backend.py
+++ b/edi_core_oca/models/edi_backend.py
@@ -10,6 +10,8 @@ import logging
 import traceback
 from io import StringIO
 
+from psycopg2 import IntegrityError, OperationalError
+
 from odoo import exceptions, fields, models
 from odoo.exceptions import UserError
 
@@ -237,6 +239,11 @@ class EDIBackend(models.Model):
             _logger.debug(
                 "%s send failed. Marked as errored.", exchange_record.identifier
             )
+        except (OperationalError, IntegrityError):
+            # We don't want the finally block to be executed in this case as
+            # the cursor is already in an aborted state and any query will fail.
+            res = "__sql_error__"
+            raise
         else:
             # TODO: maybe the send handler should return desired message and state
             message = exchange_record._exchange_status_message("send_ok")
@@ -248,16 +255,18 @@ class EDIBackend(models.Model):
             )
             res = message
         finally:
-            exchange_record.write(
-                {
-                    "edi_exchange_state": state,
-                    "exchange_error": error,
-                    "exchange_error_traceback": traceback,
-                    # FIXME: this should come from _compute_exchanged_on
-                    # but somehow it's failing in send tests (in record tests it works).
-                    "exchanged_on": fields.Datetime.now(),
-                }
-            )
+            if res != "__sql_error__":
+                exchange_record.write(
+                    {
+                        "edi_exchange_state": state,
+                        "exchange_error": error,
+                        "exchange_error_traceback": traceback,
+                        # FIXME: this should come from _compute_exchanged_on
+                        # but somehow it's failing in send tests
+                        # (in record tests it works).
+                        "exchanged_on": fields.Datetime.now(),
+                    }
+                )
         exchange_record.notify_action_complete("send", message=message)
         return res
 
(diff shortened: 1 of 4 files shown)


edi_core_oca/models/edi_backend.py
```diff
@@ -237,6 +239,11 @@ def exchange_send(self, exchange_record):
             _logger.debug(
                 "%s send failed. Marked as errored.", exchange_record.identifier
             )
+        except psycopg2.errors.InFailedSqlTransaction:
```
guewen: @Ricardoalso `InFailedSqlTransaction` is what happens when we try to run a new SQL query in an aborted transaction because of a past error, which means we already had an SQL error previously (for instance we can have an `InFailedSqlTransaction` if we do the write in the finally clause after a serialization error).
Catching this exception will not protect us against for e.g. serialization errors, because the error will pop, not be caught here, so the write in the finally will be done but fail with a `InFailedSqlTransaction`.

Now, I'm not sure what the authors would like to catch and report in the 
`exchange_record.write` of the finally clause (@simahawk @etobella ?)

At the very least, the errors from `from odoo.service.model import PG_CONCURRENCY_ERRORS_TO_RETRY` should be caught and reraised so they are retried.

Now, if we do not catch other errors in  `psycopg2.errors.OperationalError` and  `psycopg2.errors.IntegrityError`, the `write` in the finally clause will fail anyway with a `InFailedSqlTransaction`. It would be bad because it obscures the original error.

So, one option is we catch and reraise these errors too. The other option is we catch them and use a savepoint around `self._exchange_send(exchange_record)` to be able to write the error status / do the write in a new transaction.

So maybe

```python
try:
    self._exchange_send(exchange_record)
except ...:
except (OperationalError, IntegrityError):
    # if the error is retryable (one of PG_CONCURRENCY_ERRORS_TO_RETRY), it will be retried by the controller / job, otherwise the error will pop up
    _logger...
    state = "output_sql_error"
    raise
```

or

```python
try:
    with self.env.cr.savepoint():
        self._exchange_send(exchange_record)
except ...:
except PG_CONCURRENCY_ERRORS_TO_RETRY:
    _logger...
    state = "output_sql_error"
    raise
except (OperationalError, IntegrityError):
    _logger...
   # raise or not, I don't know what is supposed to happen
```

There are some variants

Ricardoalso: Indeed, thanks @guewen 

I’m more in favor of the “catch and raise” without the new transaction write approach. It prevents writing to the exchange_record, though the trade‑off is that we lose some information about the current state. However, since I see this as an intermediate state rather than a final one 🤔, it might be better to avoid creating a new transaction just to capture that temporary state.

I am open to suggestions

gurneyalex (approved):

edi_core_oca/models/edi_backend.py
```diff
@@ -237,6 +239,11 @@ def exchange_send(self, exchange_record):
             _logger.debug(
                 "%s send failed. Marked as errored.", exchange_record.identifier
             )
+        except (OperationalError, IntegrityError):
+            # We don't want the finally block to be executed in this case as
+            # the cursor is already in an aborted state and any query will fail.
+            state = "sql_error"
```
simahawk: Just to not hijack the "state" variable and make ppl think that is a valid state. Not a blocker.
```suggestion
            res = "__sql_error__"
```

edi_core_oca/models/edi_backend.py
```diff
@@ -248,16 +255,18 @@ def exchange_send(self, exchange_record):
...
-                {
-                    "edi_exchange_state": state,
-                    "exchange_error": error,
-                    "exchange_error_traceback": traceback,
-                    # FIXME: this should come from _compute_exchanged_on
-                    # but somehow it's failing in send tests (in record tests it works).
-                    "exchanged_on": fields.Datetime.now(),
-                }
-            )
+            if not state == "sql_error":
```
simahawk:
```suggestion
            if not res == "__sql_error__":
```

simahawk (commented):
Overall LGTM. Thanks!
Any chance to have a test?

Ricardoalso:
> Overall LGTM. Thanks! Any chance to have a test?

Done :pray:

simahawk (approved):
Thanks! Pls squash :)

cyrilmanuel (approved):

simahawk:
@Ricardoalso can you pls fwd port to 19?

Ricardoalso:
> @Ricardoalso can you pls fwd port to 19?

Done https://github.com/OCA/edi-framework/pull/282 :heavy_check_mark:

PR #241, opened 2026-02-25
[18.0][MIG] edi_partner_oca: Migration to 18.0 

Ricardoalso:


cyrilmanuel (approved):

santostelmo (approved):

edi_partner_oca/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "name": "EDI Partners",
+    "summary": """
+       EDI framework configuration and base logic for partners""",
+    "version": "18.0.1.0.1",
+    "license": "AGPL-3",
+    "author": "Camptocamp,Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/edi-framework",
+    "depends": [
+        "contacts",
+        "edi_endpoint_oca",
```
etobella: Why endpoint?

etobella (changes requested):
IMO edi_endpoint_oca is not required. You are adding unnnecessary requiremnts, just for adding the mixin on partner, edi_core_oca is enough.

edi_partner_oca/__manifest__.py
Ricardoalso: Indeed, it is not needed, and I have just replaced it with edi_core_oca.
If I remember correctly, I had some strange test failures in the CI back then.

The PR is now corrected with a fixup commit; let me know when it is OK with you and I will rebase.

Thanks @etobella

Ricardoalso: Okay, the same error triggered in the CI and it seems linked to dependencies load order. I will investigate

edi_partner_endpoint_oca/models/res_partner.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2026 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    origin_edi_endpoint_id = fields.Many2one(
```
simahawk, line 10: this field should appear in the EDI Tab on the partner form and optionally in the list view.

edi_partner_oca/__manifest__.py
Ricardoalso: After some moving around, I moved origin_edi_endpoint_id out of the abstract edi.exchange.consumer.mixin extension and into concrete glue modules (edi_partner_endpoint_oca and edi_sale_endpoint), then removed the old edi_endpoint_oca mixin file. This avoids global field injection and fixes the registry load-order KeyError while keeping endpoint-origin fields available where needed.

I still need to do some cleaning and commit renaming + rebases. Can I let you check the new structure proposition  ?

PR #250, opened 2026-03-25
[19.0][MIG] edi_exchange_template_oca: Migration to 19.0

sergijPf:
few changes done as per 19.0:
- replace odoo.fields.first(), removed in core;
- remove 'nocopy' param in safe_eval(), no more exists;
- odoo_test_helper.FakeModelLoader  replaced with _setup_env(), done similar way to edi_core_oca;
- remove <group> string/expand attributes for search view (edi_exchange_template_output_view_search), causing failure


edi_exchange_template_oca/models/edi_exchange_template_output.py
```diff
@@ -0,0 +1,161 @@
...
+                output = self._prettify_xml(output)
+        return output
+
+    def _post_process_output_json(self, output):
+        return json.dumps(output["payload"])
+
+    def _prettify_xml(self, xml_string):
+        return etree.tostring(etree.fromstring(xml_string), pretty_print=True)
+
+    def _get_info_provider(self, exchange_record, work_ctx=None, usage=None, **kw):
```
simahawk: @etobella this slipped out of our reviews when moving away from components: it's dead code w/o `edi_component_oca`.
In fact, this is a part that I wanted to remove anyway because the snippet already allow you to prepare data from where you want w/o needing to define a specific component.
The original idea was to be able to define default "info" components that could be reused.

@sergijPf pls get rid of it in a separate commit. Something like
```
[REM] edi_exchange_template_oca: get rid of dead code

Meant to work with components only. 
```

simahawk (changes requested):

edi_exchange_template_oca/models/edi_exchange_template_output.py
sergijPf: @simahawk removed as requested

simahawk (approved):

edi_exchange_template_oca/utils.py
```diff
@@ -0,0 +1,44 @@
+# Copyright 2020 ACSONE SA/NV
+# Copyright 2022 Camptocamp SA
+# @author Simone Orsi <simahawk@gmail.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from lxml import etree
+
+
+def xml_purge_nswrapper(xml_content):
```
simahawk, line 9: @sergijPf may I ask you you to test one thing?
The `nswrapper` trick should be superflous today because to define namespaces you can use this approach 

```
<t xmlns:cac="urn:oasis:names:specification:ubl:schema:xsd:CommonAggregateComponents-2">
            <cac:DeliveryParty>
              [...]
            </cac:DeliveryParty>
</t>
```
If this works, we can leave behind this trick.

edi_exchange_template_oca/models/edi_exchange_template_mixin.py
```diff
@@ -0,0 +1,164 @@
...
+            "uid": self.env.uid,
+            "user": self.env.user,
+            "DotDict": DotDict,
+        }
+        ctx.update(self._time_utils())
+        return ctx
+
+    def _time_utils(self):
+        def first():
+            return lambda records: next(iter(records)) if len(records) > 1 else records
```
SilvioC2C, line 125: I'm afraid this may break the code:
- `fields.first` was a function that accepted a recordset as param and returned its first record (if any)
- this new implementation, instead, accepts no arguments and returns a function

Can you test this, please?
If not working, should be easy to fix:
```
def first(recs):
    return next(iter(recs)) if len(recs) > 1 else recs
```

SilvioC2C (commented):
The PR looks good overall, I'm afraid the new `first_of` function might crash. Can you please test it (possibly add an automated test)?

PR #251, opened 2026-03-25
[18.0][FIX] ``edi_core_oca``: fix record rule

SilvioC2C:
Exchange Record's ``res_id`` is a ``Many2onReference`` field, which internally converts False-ish values to 0 before storing them to the cache and the DB.
The rule's domain old leaf ``('res_id', '=', False)`` was instead converted to a SQL query clause ``WHERE "edi_exchange_record.res_id" IS NULL``.
Since all ``edi_exchange_record`` rows contain a non-negative integer in the ``res_id`` column, the rule old domain leaf always failed to fetch any record.

Changing the leaf to ``('res_id', '=', 0)`` fixes the issue, making such Exchange Records visible again for internal users.

### DIFF
--- a/edi_core_oca/__manifest__.py
+++ b/edi_core_oca/__manifest__.py
@@ -9,7 +9,7 @@
     Define backends, exchange types, exchange records,
     basic automation and views for handling EDI exchanges.
     """,
-    "version": "18.0.1.6.3",
+    "version": "18.0.1.6.4",
     "website": "https://github.com/OCA/edi-framework",
     "development_status": "Beta",
     "license": "LGPL-3",
--- a/edi_core_oca/migrations/18.0.1.6.4/post-mig.py
+++ b/edi_core_oca/migrations/18.0.1.6.4/post-mig.py
@@ -1,0 +1,24 @@
+# Copyright 2026 Camptocamp SA (http://www.camptocamp.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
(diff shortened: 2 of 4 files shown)


edi_core_oca/__manifest__.py
```diff
@@ -9,7 +9,7 @@
     Define backends, exchange types, exchange records,
     basic automation and views for handling EDI exchanges.
     """,
-    "version": "18.0.1.6.3",
+    "version": "18.0.1.7.0",
```
simahawk: this is a patch, it should be 1.6.4

edi_core_oca/README.rst
```diff
@@ -1,7 +1,3 @@
-.. image:: https://odoo-community.org/readme-banner-image
```
simahawk, line 1: ` export SKIP=oca-gen-addon-readme`

and you'll avoid updating readme files on commit.

simahawk (changes requested):
tnx for the fix!

simahawk (approved):
@JordiMForgeFlow could you pls have a look at this?

JordiMForgeFlow (approved):
it was always a bit annoying when you had these cases, thanks for the fix!

simahawk:
@SilvioC2C can you fwd port to 19?

SilvioC2C:
> @SilvioC2C can you fwd port to 19?

@simahawk https://github.com/OCA/edi-framework/pull/271

PR #252, opened 2026-03-26
[19.0][MIG] edi_queue_oca: Migration to 19.0

sergijPf:
two major changes were done:
- method patching inside _register_hook() of edi.exchange.record has been changed according to _patch_job_auto_delay() docstring;
-  odoo_test_helper.FakeModelLoader in tests replaced with new way to load fake models (_setup_env() override);

PR #245, opened 2026-03-04
[19.0][MIG] edi_component_oca

AaronHForgeFlow:
Depends on https://github.com/OCA/connector/pull/509

PR #253, opened 2026-03-26
[19.0][MIG] edi_oca: Migration to 19.0

sergijPf:
no changes,
depends on https://github.com/OCA/edi-framework/pull/245  and https://github.com/OCA/edi-framework/pull/252


etobella:
Edi OCA will not be migrated

PR #254, opened 2026-03-26
[19.0][MIG] edi_storage_oca: Migration to 19.0

sergijPf:
- demo data removed (no more ref to storage demo in 'fs.storage');
- replace demo data references in tests;


simahawk:
to be included https://github.com/OCA/edi-framework/pull/280

acsonefho (approved):

ivantodorovich (approved):

etobella (approved):

edi_storage_oca/demo/edi_backend_demo.xml
```diff
@@ -1,14 +0,0 @@
-<?xml version="1.0" encoding="utf-8" ?>
```
simahawk, line 1: I see this approach a bit everywhere: deleting demo data because they are not use in tests.
1. they could still be used in tests by forced load
2. they could still be kept for.... demoing!
Not a blocker, but just a reminder such cleanups are not mandatoy 😉

simahawk (approved):

PR #255, opened 2026-03-27
[19.0][MIG] edi_stock_oca: Migration to 19.0

sergijPf:
- replace module dependencies (edi_oca => edi_core_oca, component_event => edi_component_oca)
edi_oca is deprecated https://github.com/OCA/edi-framework/pull/253
- replace tests' demo data


edi_stock_oca/components/listeners.py
```diff
@@ -0,0 +1,24 @@
+from odoo.addons.component.core import Component
+
+
+class EDIConfigSPListener(Component):
```
simahawk, line 4: Not a blocker but we can now get rid of listeners and rely only on edi.conf records.
Shall we take the occasion?

edi_stock_oca/models/stock_picking.py
```diff
@@ -0,0 +1,42 @@
...
+from contextlib import contextmanager
+
+from odoo import models
+
+
+class StockPicking(models.Model):
+    _name = "stock.picking"
+    _inherit = ["stock.picking", "edi.exchange.consumer.mixin"]
+
+    # Override all methods to trigger EDI exchange after picking state changes.
```
simahawk, line 14: we could get rid of these methods

simahawk (changes requested):
@etobella ping :)

edi_stock_oca/__manifest__.py
```diff
@@ -5,11 +5,11 @@
     "name": "EDI Stock OCA",
     "summary": """
        Define EDI Configuration for Stock""",
-    "version": "18.0.1.0.1",
+    "version": "19.0.1.0.0",
     "license": "AGPL-3",
     "author": "Creu Blanca, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/edi-framework",
-    "depends": ["stock", "edi_oca", "component_event"],
```
etobella, line 12: For me, now is the moment to split components from edi_stock. WDYT?

If components are required, we can made edi_stock_component.

PR #257, opened 2026-04-01
[17.0][FIX] edi_record_metadata_oca: rename get_metadata (ORM conflict)

vvrossem:
`get_metadata()` is a built-in Odoo ORM method called by the "Debug > View Metadata" action. Defining it on `edi.exchange.record` overrode that behavior, causing a JS crash when opening the debug dialog:

  "Cannot read properties of undefined (reading 'id')"

Renamed to `edi_get_metadata()` / `edi_set_metadata()` (for consistency) and updated all callers.

PR #258, opened 2026-04-01
[18.0][FIX] edi_record_metadata_oca: rename get_metadata (ORM conflict)

vvrossem:
`get_metadata()` is a built-in Odoo ORM method called by the "Debug > View Metadata" action. Defining it on `edi.exchange.record` overrode that behavior, causing a JS crash when opening the debug dialog:

  "Cannot read properties of undefined (reading 'id')"

Renamed to `edi_get_metadata()` / `edi_set_metadata()` (for consistency) and updated all callers.


SilvioC2C (approved):
LGTM

Please squash the commits or simply use `git cherry-pick` to forward-port commits from #257

simahawk (approved):

PR #259, opened 2026-04-02
[19.0][MIG] edi_record_metadata_oca: Migration to 19.0

sergijPf:
regular migration, no changes except FakeModeLoader in tests, replaced with new approach


edi_record_metadata_oca/models/edi_exchange_record.py
```diff
@@ -0,0 +1,32 @@
...
+
+from odoo import api, fields, models
+
+from ..fields import BetterSerialized
+
+
+class EDIExchangeRecord(models.Model):
+    _inherit = "edi.exchange.record"
+
+    metadata = BetterSerialized(
```
ivantodorovich: I think we can use the rather new (a few version ago) `fields.Json` instead, and drop the `base_sparse_field` dependency and the custom "BetterSerialized" field.

ivantodorovich (approved):

yankinmax:
Hello @sergijPf can you pls include this improvement:
- [ ] https://github.com/OCA/edi-framework/pull/258

yankinmax (approved):

edi_record_metadata_oca/models/edi_exchange_record.py
sergijPf: @ivantodorovich thanks for review, changed as suggested,
'base_sparse_field' removed from dependencies (it still to be installed because 'edi_core_oca' depends on it)

sergijPf:
> Hello @sergijPf can you pls include this improvement:
> 
>     * [ ]  [[18.0][FIX] edi_record_metadata_oca: rename get_metadata (ORM conflict) #258](https://github.com/OCA/edi-framework/pull/258)

@yankinmax thanks for review, it was already considered originally, the only thing, I've updated inside readme.rst too

edi_record_metadata_oca/models/edi_exchange_record.py
ivantodorovich: Thanks!

Don't we need a migration script for the `metadata` field type change?
AFAIK `Serielized` is stored as a string, whilst `Json` is stored as JSONB. I don't think the ORM takes care of this conversion, but maybe I'm wrong

sergijPf: no problem with it as per local test:
before:
image
after:
image
image

grindtildeath (approved):

yankinmax:
Hello @OCA/edi-framework-maintainers can you pls take a look and possibly merge?

vvrossem (approved):
Migration commit LG

edi_record_metadata_oca/fields.py
```diff
@@ -29,17 +27,3 @@ def convert_datetime(obj):
 class ExtendedJSONEncoder(json.JSONEncoder):
```
simahawk: if we do not use this class.... why do we keep this file? 🤔 
And... are we sure that dates are converted properly by the std field?

edi_record_metadata_oca/tests/test_metadata.py
```diff
@@ -2,34 +2,39 @@
...
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
-from odoo_test_helper import FakeModelLoader
-
 from odoo import fields
+from odoo.orm.model_classes import add_to_registry
 
 from odoo.addons.edi_core_oca.tests.common import EDIBackendCommonTestCase
 
+from .fake_models import EDIMetadataConsumerFake
```
simahawk: the import of the test model should be done inside `_setup_env`

simahawk (changes requested):

simahawk:
@sergijPf gentle ping :)

edi_record_metadata_oca/tests/test_metadata.py
sergijPf: moved

edi_record_metadata_oca/fields.py
sergijPf: 1) class removed.
2) I did simple check with date, and it seems to be ok:
image

not sure with other possible cases, it'll be better to test it with real (live) example dataset,
I don't have such, and didn't use this module before (migrated it only because of edi_sale dependency)

simahawk: I was referring to the entire file. The `convert` function is not used anymore.
Plus, we have tests for dates conversion (I'm glad I did that :smile: ) -> it's working w/o that already.
You can just trash the file :wink:

sergijPf: removed

simahawk (approved):

PR #261, opened 2026-04-06
[19.0] edi_core_oca: remove demo test data

yankinmax:
> If your tests rely on demo data, now they are not installed by default, as in fact, Odoo advises not to rely on them, so it's the occasion to create the data on the tests themselves.

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-19.0

The demo data is actually created in tests, so I think it's a leftover, but it causes tests fail on other migration (duplicate key for the backend type).
https://github.com/OCA/edi-framework/blob/e9eeec0cea3cdf49be69da0733a238c3e4ade666/edi_core_oca/tests/common.py#L68-L83


yankinmax:
It seems I've wrongly opened this PR and another modules tests are OK.
This is a usual DEMO data for the module.

PR #272, opened 2026-05-06
[19.0] [IMP] edi_endpoint_oca: exec_mode=create_exchange_record

ivantodorovich:
Added a new default exec_mode for EDI exchange records: `create_exchange_record`

This mode will dump the request body as-is into the exchange record content for a later processing.

### DIFF
--- a/edi_endpoint_oca/models/edi_endpoint.py
+++ b/edi_endpoint_oca/models/edi_endpoint.py
@@ -66,6 +66,7 @@ class EDIEndpoint(models.Model):
             if not isinstance(file_content, bytes):
                 file_content = bytes(file_content, encoding)
             vals["exchange_file"] = base64.b64encode(file_content)
+            vals["edi_exchange_state"] = "input_received"
 
         rec = self.backend_id.create_record(self.exchange_type_id.code, vals)
         return rec
--- a/edi_endpoint_oca/models/edi_endpoint.py
+++ b/edi_endpoint_oca/models/edi_endpoint.py
@@ -33,6 +33,26 @@ class EDIEndpoint(models.Model):
         comodel_name="edi.exchange.type",
         domain="[('backend_type_id','=', backend_type_id)]",
     )
+    exec_mode = fields.Selection(default="create_exchange_record")
+
+    def _selection_exec_mode(self):
+        return super()._selection_exec_mode() + [
+            ("create_exchange_record", self.env._("Create exchange record")),
+        ]
+
+    def _handle_exec__create_exchange_record(self, request):
+        """Persist the raw HTTP body as an exchange record and acknowledge.
+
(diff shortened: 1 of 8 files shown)


edi_endpoint_oca/models/edi_endpoint.py
```diff
@@ -33,6 +33,26 @@ class EDIEndpoint(models.Model):
...
+
+    def _handle_exec__create_exchange_record(self, request):
+        """Persist the raw HTTP body as an exchange record and acknowledge.
+
+        Covers the "receive and queue" case for incoming EDI endpoints,
+        avoiding the need for a per-endpoint code snippet.
+        """
+        record = self.create_exchange_record(
+            file_content=request.httprequest.get_data(),
+        )
```
ivantodorovich, line 51: @simahawk shouldn't `create_exchange_record` set the edi_exchange_state to `input_received` if `file_content` is provided?
or perhaps it should immediately call exchange_receive in such case?

To me it seems obvious, I don't get why we would want it in a separate step given the file_content is already set.

simahawk: makes sense

edi_endpoint_oca/models/edi_endpoint.py
```diff
@@ -33,6 +33,26 @@ class EDIEndpoint(models.Model):
         comodel_name="edi.exchange.type",
         domain="[('backend_type_id','=', backend_type_id)]",
     )
+    exec_mode = fields.Selection(default="create_exchange_record")
+
+    def _selection_exec_mode(self):
+        return super()._selection_exec_mode() + [
+            ("create_exchange_record", self.env._("Create exchange record")),
```
simahawk, line 40: there's no need to translate selection items. Is it different now?

simahawk (approved):
LG

vvrossem (approved):
🐐

edi_endpoint_oca/models/edi_endpoint.py
ivantodorovich: Done!

edi_endpoint_oca/models/edi_endpoint.py
ivantodorovich: That's only true for pure `selection=list`, but not when you define a method that builds the options.
See: https://github.com/odoo/odoo/blob/3ced6eb35bc2617310c4413190a35e3b08617dca/odoo/orm/fields_selection.py#L204-L207

If the selection is defined by a method, it's the method's responsibility to translate the labels.

simahawk:
/ocabot merge minor

@ivantodorovich can you pls backport this to 18?

ivantodorovich:
> can you pls backport this to 18?

Backport PR:
- https://github.com/OCA/edi-framework/pull/281
- https://github.com/OCA/web-api/pull/132

PR #273, opened 2026-05-12
[19.0][IMP] edi_*_oca: skip exchange types w/o handler in cron domains

HviorForgeFlow:
Noticed while developing that when a exchange type is not linked to a handler and the EDINotImplementedError is raised, a DB rollback takes places undoing all previous work. 

To ensure that skipping not handled exchanges  does not break anything outside the cron scope I had to slightly change the exchange_generate_send method.

CC @etobella @simahawk @ForgeFlow

### DIFF
--- a/edi_core_oca/models/edi_backend.py
+++ b/edi_core_oca/models/edi_backend.py
@@ -403,6 +403,8 @@ class EDIBackend(models.Model):
             ("backend_id", "=", self.id),
             ("type_id.exchange_file_auto_generate", "=", True),
             ("type_id.direction", "=", "output"),
+            ("type_id.generate_model_id", "!=", False),
+            ("type_id.send_model_id", "!=", False),
             ("edi_exchange_state", "=", "new"),
             ("exchange_file", "=", False),
         ]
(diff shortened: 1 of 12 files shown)


simahawk:
Wondering: why is it possible to handle an exc type w/o handler? It should be validated beforehand...

We should make specific validators required based on the direction.
And maybe improve `_$direction_new_records_domain`.

HviorForgeFlow:
> Wondering: why is it possible to handle an exc type w/o handler? It should be validated beforehand...
> 
> We should make specific validators required based on the direction. And maybe improve `_$direction_new_records_domain`.

That makes more sense, cleaner implementation, thanks for the advice

HviorForgeFlow:
@simahawk it follows your suggested approach

JordiMForgeFlow (approved):

AaronHForgeFlow (approved):

MiquelRForgeFlow (approved):

simahawk:
@HviorForgeFlow I think you miss the part "specific validators required based on the direction."
Now that I see the change I'm not really a big fan of just the domain fix. Sorry :innocent: 

What about adding validation and fixing the data the you have?
I mean, an exc type that has direction "output" and has no "generate" or "send" handler is definitely broken.
I'd rather:
* add validation on the exc type form (make some fields required based on direction)
* add an api.constraint

WDYT?

simahawk:
@HviorForgeFlow ping :)

HviorForgeFlow:
Proposal on adding validators and constraints to the exchange type as you were suggesting @simahawk

MiquelRForgeFlow:
@simahawk could you check?

simahawk (approved):

PR #275, opened 2026-05-14
[18.0][IMP] edi_core: add global events handling via edi.conf

simahawk:
Replaces #232 
See atomic commits and updated docs for details :wink: 
Long story short: we now have global events, that means we can avoid using component_event if not required.

### DIFF
--- a/edi_component_oca/models/edi_exchange_record.py
+++ b/edi_component_oca/models/edi_exchange_record.py
@@ -10,7 +10,7 @@ class EdiExchangeRecord(models.Model):
 
     def _trigger_edi_event(self, name, suffix=None, target=None, **kw):
         """Trigger a component event linked to this backend and edi exchange."""
-        name = self._trigger_edi_event_make_name(name, suffix=suffix)
+        event_name = self._trigger_edi_event_make_name(name, suffix=suffix)
         target = target or self
-        target._event(name).notify(self, **kw)
+        target._event(event_name).notify(self, **kw)
         return super()._trigger_edi_event(name, suffix=suffix, target=target, **kw)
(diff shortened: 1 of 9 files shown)


simahawk:
@ArnauCForgeFlow @etobella could you please have a look at this? :innocent:

edi_component_oca/models/edi_exchange_record.py
```diff
@@ -8,12 +8,6 @@
 class EdiExchangeRecord(models.Model):
     _inherit = "edi.exchange.record"
 
-    def _trigger_edi_event_make_name(self, name, suffix=None):
-        return "on_edi_exchange_{name}{suffix}".format(
-            name=name,
-            suffix=("_" + suffix) if suffix else "",
-        )
-
     def _trigger_edi_event(self, name, suffix=None, target=None, **kw):
```
yankinmax, line 11: if `super` receives `event_name`, core expands it again into `on_edi_exchange_on_edi_exchange_done`. Passing raw `name` lets core produce the correct `on_edi_exchange_done`.
```
def _trigger_edi_event(self, name, suffix=None, target=None, **kw):
    """Trigger a component event linked to this backend and edi exchange."""
    event_name = self._trigger_edi_event_make_name(name, suffix=suffix)
    target = target or self
    target._event(event_name).notify(self, **kw)
    return super()._trigger_edi_event(
        name, suffix=suffix, target=target, **kw
    )
```

yankinmax (approved):
Thank you @simahawk

edi_component_oca/models/edi_exchange_record.py
simahawk: ah, good catch 👍

ArnauCForgeFlow (approved):
LGTM. Reimplemented **edi_storage_oca** done/error listener on top of this, works nicely

ivantodorovich (approved):
LG

PR #280, opened 2026-05-19
[18.0][IMP] edi_storage_oca: post-process input file via conf

ArnauCForgeFlow:
Reintroduce the event-driven file move "on_edi_exchange_done" / "on_edi_exchange_error" that was lost when the "edi_component_oca" dependency was dropped, using the global edi.configuration mechanism added to "edi_core_oca" in #275:

  - Add `_move_file` on `fs.storage`: looks up the file in the source directory and schedules an `fs.move` via a post-commit hook.
  - Add `_storage_on_edi_exchange_done` / `_storage_on_edi_exchange_error` on `edi.backend`, mirroring the original listener behaviour.
  - Ship two default global `edi.configuration` records bound to the matching triggers. They are inactive by default and require an explicit `backend_id` before being enabled.

Depends on:
- [x] #275

@simahawk

### DIFF
--- a/edi_storage_oca/data/edi_configuration.xml
+++ b/edi_storage_oca/data/edi_configuration.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
+    <!--
+        IMPORTANT: these configurations are *inactive* by default. Before
+        activating them, set 'backend_id' to the specific storage backend
+        whose directories should be used, otherwise the snippet will not
+        know which input_dir_pending / input_dir_done / input_dir_error to
+        operate on.
+    -->
+    <record id="edi_conf_storage_move_on_done" model="edi.configuration">
+        <field name="name">Storage: move input file to done</field>
+        <field name="active" eval="False" />
+        <field name="is_global" eval="True" />
+        <field name="trigger_id" ref="edi_core_oca.edi_config_trigger_record_done" />
+        <field
+            name="snippet_do"
+        >record.backend_id._storage_on_edi_exchange_done(record)</field>
+    </record>
+
+    <record id="edi_conf_storage_move_on_error" model="edi.configuration">
+        <field name="name">Storage: move input file to error</field>
+        <field name="active" eval="False" />
+        <field name="is_global" eval="True" />
+        <field name="trigger_id" ref="edi_core_oca.edi_config_trigger_record_error" />
+        <field
+            name="snippet_do"
+        >record.backend_id._storage_on_edi_exchange_error(record)</field>
+    </record>
+</odoo>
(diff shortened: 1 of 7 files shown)


edi_storage_oca/data/edi_configuration.xml
```diff
@@ -0,0 +1,35 @@
...
+        <field name="name">Storage: move input file to done</field>
+        <field name="active" eval="False" />
+        <field name="is_global" eval="True" />
+        <field name="trigger_id" ref="edi_core_oca.edi_config_trigger_record_done" />
+        <field
+            name="snippet_do"
+        >record.backend_id._storage_on_edi_exchange_done(record)</field>
+        <field
+            name="description"
+        >Pick a backend before enabling: snippet uses backend.input_dir_pending/done/error.</field>
```
simahawk: this should describe what it does. This message is more for the snippet of code

edi_storage_oca/models/fs_storage.py
```diff
@@ -0,0 +1,28 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+import functools
+import os.path
+from pathlib import PurePath
+
+from odoo import models
+
+
+class FsStorage(models.Model):
```
simahawk: TODO: move to `fs_storage` module

edi_storage_oca/static/description/index.html
```diff
@@ -3,7 +3,7 @@
 <head>
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
 <meta name="generator" content="Docutils: https://docutils.sourceforge.io/" />
-<title>README.rst</title>
```
simahawk, line 6: `export SKIP=oca-gen-addon-readme` to disable readme updates

edi_storage_oca/tests/test_event_listener.py
```diff
@@ -0,0 +1,89 @@
...
+from odoo.addons.edi_core_oca.tests.common import EDIBackendCommonTestCase
+
+STORAGE_MOVE_FILE_PATH = (
+    "odoo.addons.edi_storage_oca.models.fs_storage.FsStorage._move_file"
+)
+
+
+class TestStorageEventListener(EDIBackendCommonTestCase):
+    @classmethod
+    def setUpClass(cls):
```
simahawk, line 23: I'm not sure the fake model setup works anymore at `setUpClass`.
You can keep the rest here and move just the model setup to `setUp`.

simahawk (changes requested):

ArnauCForgeFlow:
Rebased!

simahawk:
@ArnauCForgeFlow I forgot that we already have utils in edi_storage to deal w/ files. Since the discussion on https://github.com/OCA/storage/pull/606 will takel a while I would do this: you could improve https://github.com/OCA/edi-framework/blob/18.0/edi_storage_oca/utils.py#L56 for now and add a TODO to followup w/ that PR. Does it sounds good for you?

ArnauCForgeFlow:
Hi @simahawk For now, I added a new  method called `move_file` inside utils, with a TODO comment that it should be dropped once https://github.com/OCA/storage/pull/606 is merged.

ArnauCForgeFlow:
@simahawk kind reminder!

simahawk (approved):

PR #287, opened 2026-05-22
[19.0][MIG] edi_product_oca: Migration to 19.0

yankinmax:

PR #290, opened 2026-05-25
[18.0][FIX] edi_product_oca: fix wrong dep on edi_endpoint_oca

yankinmax:
See requested changes:
- [ ] https://github.com/OCA/edi-framework/pull/287#discussion_r3296830201

cc @simahawk

### DIFF
--- a/edi_product_oca/__manifest__.py
+++ b/edi_product_oca/__manifest__.py
@@ -11,7 +11,14 @@
     "author": "ForgeFlow, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/edi-framework",
     "depends": [
+        # Odoo/core
         "product",
+        # OCA/edi-framework
+        # TODO: Replace this dependency with ``edi_core_oca`` in version 19.0
+        # For version 18.0 without ``edi_endpoint_oca``, CI fails
+        # because ``origin_edi_endpoint_id`` is set up on product models
+        # while its related target ``edi.exchange.record.edi_endpoint_id`` is absent.
+        # A glue module similar to ``edi_sale_endpoint`` does not currently solve it.
         "edi_endpoint_oca",
     ],
     "data": [


simahawk:
the reason is not "edi core revamp". The dependency on endpoint module is wrong by itself and I don't know why it has been added :wink:

yankinmax:
@simahawk good to know, thanks

simahawk:
I'm checking the test issue

cyrilmanuel (approved):

simahawk:
@yankinmax I've not managed to fix this yet. Do me a favor: live it like it is today w/ the dependency but add a TODO and report there the motivation. A simple glue module for edi_product + endpoint does not work as it does for edi_sale_endpoint so... let's not waste too much time on something that is like that since ages.
Then apply the same comment on v19 and we are done.
Thanks!

yankinmax:
@simahawk done!

simahawk:
@yankinmax pls rewrite the commit msg

edi_product_oca/__manifest__.py
```diff
@@ -11,7 +11,14 @@
...
     "website": "https://github.com/OCA/edi-framework",
     "depends": [
+        # Odoo/core
         "product",
+        # OCA/edi-framework
+        # TODO: Replace this dependency with ``edi_core_oca`` in version 19.0
+        # For version 18.0 without ``edi_endpoint_oca``, CI fails
+        # because ``origin_edi_endpoint_id`` is set up on product models
+        # while its related target ``edi.exchange.record.edi_endpoint_id`` is absent.
+        # A glue module similar to ``edi_sale_endpoint`` does not currently solve it.
```
SilvioC2C, line 21:
```suggestion
        # Deprecated in favor of ``edi_core_oca`` in version 19.0
        # In 18.0, CI would fail a direct dependency on ``edi_endpoint_oca``,
        # because ``origin_edi_endpoint_id`` is set up on products
        # while its related target ``edi.exchange.record.edi_endpoint_id`` is absent.
        # A glue module similar to ``edi_sale_endpoint`` does not currently solve it.
```

SilvioC2C (approved):
LGTM, I proposed a minor improvement to the comment (since we already have removed that module dependency in v19 😉)

PR #292, opened 2026-05-28
[18.0][IMP] edi_core_oca: type add action_view_exchange_records

simahawk:
You can finally jump from a type to its existing records.

### DIFF
--- a/edi_core_oca/models/edi_exchange_type.py
+++ b/edi_core_oca/models/edi_exchange_type.py
@@ -147,6 +147,14 @@ class EDIExchangeType(models.Model):
         """,
     )
     advanced_settings = Serialized(default={}, compute="_compute_advanced_settings")
+    exchange_record_ids = fields.One2many(
+        comodel_name="edi.exchange.record",
+        inverse_name="type_id",
+    )
+    exchange_record_count = fields.Integer(
+        string="# Exchange Records",
+        compute="_compute_exchange_record_count",
+    )
     rule_ids = fields.One2many(
         comodel_name="edi.exchange.type.rule",
         inverse_name="type_id",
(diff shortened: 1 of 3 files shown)


edi_core_oca/models/edi_exchange_type.py
```diff
@@ -240,6 +248,16 @@ def _compute_ack_for_type_ids(self):
         for rec in self:
             rec.ack_for_type_ids = [x.id for x in by_type_id.get(rec.id, [])]
 
+    def _compute_exchange_record_count(self):
```
SilvioC2C, line 252: Please add `@api.depends("exchange_record_ids")` to correctly handle the recomputation of this field

edi_core_oca/tests/test_exchange_type.py
```diff
@@ -167,3 +167,41 @@ def test_archive_rules(self):
...
+        )
+
+    def test_exchange_record_count(self):
+        exc_type = self.exchange_type_out
+        self.assertEqual(exc_type.exchange_record_count, 0)
+        rec1 = self._create_exchange_record(exc_type)
+        rec2 = self._create_exchange_record(exc_type)
+        # Record on a different type must not be counted
+        self._create_exchange_record(self.exchange_type_in)
+        exc_type.invalidate_recordset(["exchange_record_count"])
```
SilvioC2C: No need to manually invalidate the cache once the dependencies are defined correctly on the compute method

SilvioC2C (changes requested):

edi_core_oca/models/edi_exchange_type.py
simahawk: we don't need it because it might trigger useless computation for a nr that is available only on a single record view

edi_core_oca/tests/test_exchange_type.py
simahawk: see answer above

edi_core_oca/models/edi_exchange_type.py
SilvioC2C: The recomputation is triggered only when the field is read and its value can't be found in cache, so no risk of useless recomputations; on the other hand, without the deps, the cached value is never invalidated, which means that you need to manually invalidate the cache to prevent the risk of reading an outdated value (which is why tests fail without manually calling `invalidate_recordset()`).

SilvioC2C (approved):
LGTM!

etobella (approved):
Makes sense

/ocabot merge patch

PR #294, opened 2026-06-03
[16.0] [FIX] edi_oca: search with slice only can be count=False

luisDIXMIT:
If we want to slice the search method result, it must always be a list, not an integer, which is what the search method returns when count=True. Sometimes count=True is propagated and when the workflow reaches this point, it raises an exception like this:

```
File "/opt/odoo/auto/addons/edi_oca/models/edi_exchange_record.py", line 627, in _search
self._search(
TypeError: 'int' object is not subscriptable
```

### DIFF
--- a/edi_oca/models/edi_exchange_record.py
+++ b/edi_oca/models/edi_exchange_record.py
@@ -629,7 +629,7 @@ class EDIExchangeRecord(models.Model):
                     offset=offset + len(orig_ids),
                     limit=limit,
                     order=order,
-                    count=count,
+                    count=False,
                     access_rights_uid=access_rights_uid,
                 )[: limit - len(result)]
             )


etobella (approved):
It makes sense

SilvioC2C (approved):

luisDIXMIT:
Can be this PR be merged?

simahawk (approved):

simahawk:
Should it be ported to higher versions?
/ocabot merge patch

luisDIXMIT:
> Should it be ported to higher versions? /ocabot merge patch

Hi @simahawk , i have checked it. In version 16, _search method can be return a list or an integer, depends on count parameter. In version 17 and later it returns a query object and the parameter count doesn't exists, so I think it isn't necesary.

https://github.com/odoo/odoo/blob/c28ce84dd753c7f43aa21f09a72aae0fa1215754/odoo/models.py#L5416

PR #295, opened 2026-06-04
[19.0][ADD] edi_error_activities: implement error activity handling on top of edi.configuration

Ricardoalso:
This module provides generic, configurable error-activity handling for EDI exchanges.

It extends `edi.configuration` with options to control:

- whether activities are created when exchanges fail,
- which exchange states are considered errors,
- optional filtering by target model,
- assignee strategy (including optional fixed user),
- activity type,
- deduplication marker template.

It extends `edi.exchange.record` with reusable helpers that:

- evaluate whether an exchange state should trigger an activity,
- compute assignee and deduplication marker values,
- schedule deduplicated activities,
- post a message including a link to the exchange record.

This module is backend-agnostic and can be reused by any EDI integration.
Business-specific configuration records and snippets should be implemented in integration modules.


simahawk:
Do you know that we have this? https://github.com/OCA/edi-framework/tree/18.0/edi_notification_oca

simahawk (changes requested):
🔴

Ricardoalso:
Reinventing the wheel 🥲 Thanks @simahawk

Closing this one

PR #296, opened 2026-06-04
[19.0][MIG] edi_notification_oca: Migration to 19.0

Ricardoalso:


edi_notification_oca/models/edi_exchange_record.py
```diff
@@ -7,3 +7,39 @@
 class EDIExchangeRecord(models.Model):
     _name = "edi.exchange.record"
     _inherit = ["edi.exchange.record", "mail.activity.mixin"]
+
+    def _notify_error(self, message_key):
+        res = super()._notify_error(message_key)
+        if message_key == "process_ko":
```
simahawk: this change to get rid of the component dep must go in its own commit.
Also the proper way would be to use global edi.conf.
I would do it roughly like this:

* move this piece of code to a new abstract model `edi.notification.tool`
* define a new global edi.configuration that uses `edi_core_oca.edi_config_trigger_record_error` as trigger
* call `env["edi.notification.tool"].on_edi_exchange_error(exchange_record)`

simahawk (changes requested):

edi_notification_oca/models/edi_exchange_record.py
simahawk: Also note that the conditions checked in the code could be put in the before do snippet (Eg: the check for `notify_on_process_error` flag on the type). That could set the `todo` flag to false, causing the non execution of the snippet_do. However, this is a detail, maybe you can just add a TODO for now.

Ricardoalso: This idea around the `snippet_before_do` kind of pleases me 

Does the following check seems relevant to you ? 

```
bool(
  exc_type.notify_on_process_error
  and exc_type.notify_on_process_error_activity_type_id
  and (
      exc_type.notify_on_process_error_groups_ids
      or exc_type.notify_on_process_error_users_ids
  )
)
```

simahawk: yeah, the same conditions that you have in the listener now.
To simplify, we can add a method to the type `_edi_notify_on_error_enabled` something like that.

ivantodorovich (approved):
Thanks!

edi_notification_oca/data/edi_configuration.xml
```diff
@@ -0,0 +1,21 @@
...
+        <field name="is_global">True</field>
+        <field name="trigger_id" ref="edi_core_oca.edi_config_trigger_record_error" />
+        <field name="snippet_before_do">
+exc_type = record.type_id
+result = {
+    "todo": exc_type._edi_notify_on_error_enabled()
+}
+        </field>
+        <field name="snippet_do">
+env = record.env
```
simahawk: `env` is already available. If it is to preserve the record context ok but the env.context should already contain the same stuff

edi_notification_oca/models/edi_exchange_type.py
```diff
@@ -32,6 +32,17 @@ class EDIExchangeType(models.Model):
         default=lambda self: self._default_notify_on_process_error_activity_type_id(),
     )
 
+    def _edi_notify_on_error_enabled(self):
```
simahawk: I've just realized this could be a compute field (that can be displayed on the UI if needed).
Not a blocker

edi_notification_oca/models/edi_notification_tool.py
```diff
@@ -0,0 +1,32 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import models
+
+
+class EDINotificationTool(models.AbstractModel):
+    _name = "edi.notification.tool"
+    _description = "EDI Notification Tool"
+
+    def on_edi_exchange_error(self, exchange_record):
```
simahawk, line 11: a docstring would be nice and in particular can describe that it can be used by edi.configuration bla blab

edi_notification_oca/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,10 @@
+This module creates activities for users when an exchange record's process fails.
```
simahawk, line 1: Shall we describe the usage of edi.configuration in a configure file ?

simahawk (changes requested):

edi_notification_oca/models/edi_exchange_type.py
Ricardoalso: Makes sense :+1:

edi_notification_oca/data/edi_configuration.xml
Ricardoalso: It is actually not :thinking: 

CI is complaining : 
```

ValueError: NameError("name 'env' is not defined") while evaluating
'\nresult = env["edi.notification.tool"].on_edi_exchange_error(record)\n        '
```

[I will propose an improvement in edi_core_oca](https://github.com/OCA/edi-framework/pull/302) :wink:

simahawk: image

simahawk (approved):
pls squash

ivantodorovich:
should be ok to merge now

PR #298, opened 2026-06-05
[19.0][FIX] edi_core_oca: fire on_edi_exchange_error for all error states

Ricardoalso:
When using edi_notification_oca, EDI exchanges that failed during send, receive, or validation did not trigger any notification, leaving errors unnoticed and requiring manual detection.

The backend only emitted "on_edi_exchange_error" on the process error path. Other error paths updated the state and posted chatter messages but did not fire the event.

This change ensures the error event is emitted on all error transitions:

exchange_send -> _notify_error("send_ko")

exchange_receive -> _notify_error("receive_ko", "validate_ko")

exchange_generate -> _notify_error("validate_ko")

Calls are guarded by state changes to avoid duplicate notifications. Redundant chatter messages are removed since _notify_error already handles logging and event emission.

### DIFF
--- a/edi_core_oca/models/edi_backend.py
+++ b/edi_core_oca/models/edi_backend.py
@@ -121,6 +121,7 @@ class EDIBackend(models.Model):
         :param kw: keyword args to be propagated to output generate handler
         """
         self.ensure_one()
+        old_state = exchange_record.edi_exchange_state
         if force and exchange_record.exchange_file:
             # Remove file to regenerate
             exchange_record.exchange_file = False
@@ -148,7 +149,6 @@ class EDIBackend(models.Model):
                 traceback = _get_exception_traceback()
                 error = _get_exception_msg(err)
                 state = "validate_error"
-                message = exchange_record._exchange_status_message("validate_ko")
                 exchange_record.update(
                     {
                         "edi_exchange_state": state,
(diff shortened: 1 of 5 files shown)


edi_core_oca/models/edi_backend.py
```diff
@@ -281,6 +289,8 @@ def exchange_send(self, exchange_record):
                         "exchanged_on": fields.Datetime.now(),
                     }
                 )
+                if old_state != state and state == "output_error_on_send":
```
simahawk, line 292: this way you trigger it only if the error is of sql type. Am I wrong?

edi_core_oca/models/edi_backend.py
```diff
@@ -561,6 +569,10 @@ def exchange_receive(self, exchange_record):
                         "exchanged_on": fields.Datetime.now(),
                     }
                 )
+                if old_state != state and state == "input_receive_error":
```
simahawk, line 572: same

simahawk (approved):
Good catch! Thanks!

edi_core_oca/models/edi_backend.py
simahawk: I'm wrong... it's a `!=` :facepalm:

simahawk:
can you please backport to v18?
/ocabot merge patch

PR #300, opened 2026-06-05
[18.0][FIX] edi_core_oca: fix ``active_test`` propagation

SilvioC2C:
Exchange type action ``act_open_edi_exchange_type_view`` declared ``active_test=False`` in its context to allow displaying archived rules in the ExcType form view. However, this choice had the side effect of propagating ``active_test=False`` to all relational fields' context, leading to archived records to *always* be displayed. This fix removes the context key from the action and moves it to the field's definition itself (at model lever, not view level).


simahawk (approved):

simahawk:
/ocabot merge patch

To be fwd ported to 19

PR #301, opened 2026-06-05
[18.0][IMP] edi_core_oca: Auto cancel exchange records that cannot be sent

grindtildeath:
This allows to set outgoing exchange records as canceled in case they cannot be sent by customizing edi.backend._output_check_send.

### DIFF
--- a/edi_core_oca/models/edi_backend.py
+++ b/edi_core_oca/models/edi_backend.py
@@ -62,6 +62,9 @@ class EDIBackend(models.Model):
     Usecase: the web service you send the file to processes it on the fly.
     """
     )
+    output_cancel_cannot_send_auto = fields.Boolean(
+        string="Automatically cancel the records that cannot be sent"
+    )
     active = fields.Boolean(default=True)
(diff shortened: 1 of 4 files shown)


edi_core_oca/models/edi_backend.py
```diff
@@ -225,7 +228,11 @@ def exchange_send(self, exchange_record):
         # In case already sent: skip sending and check the state
         check = self._output_check_send(exchange_record)
         if not check:
-            return self._failed_output_check_send_msg()
+            message = self._failed_output_check_send_msg(exchange_record=exchange_record)
```
simahawk, line 231: This would work only for exchanges implementing `check` handlers.
Is that what you want?

simahawk (changes requested):
what is your final goal / use case?

simahawk:
Maybe this is what you are looking for https://github.com/OCA/edi-framework/pull/340

grindtildeath:
IIRC this is now solved with on fail hooks, so we can close this

PR #304, opened 2026-06-10
[19.0][IMP] edi_exchange_deduplicate_oca:  add configurable deduplication states

wouitmil:
Implement deduplicate_on_exchange_record_status on Exchange Record type setting that allows you to set what states should trigger the obsolescence of an exchange record

### DIFF
--- a/edi_exchange_deduplicate_oca/models/edi_exchange_type.py
+++ b/edi_exchange_deduplicate_oca/models/edi_exchange_type.py
@@ -1,7 +1,8 @@
 # Copyright 2024 Camptocamp
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class EDIExchangeType(models.Model):
@@ -18,3 +19,35 @@ class EDIExchangeType(models.Model):
         default=True,
         help="Delete records marked as obsolete.",
     )
+
+    deduplicate_on_exchange_record_status = fields.Char(
+        default="new,output_pending",
+        groups="base.group_no_one",
+    )
+
+    def _deduplicate_get_exchange_record_states(self):
+        self.ensure_one()
+        return {
+            state.strip()
+            for state in (self.deduplicate_on_exchange_record_status or "").split(",")
+            if state.strip()
+        }
+
+    @api.constrains("deduplicate_on_exchange_record_status")
+    def _check_deduplicate_on_exchange_record_status(self):
+        exchange_state_field = self.env["edi.exchange.record"]._fields[
+            "edi_exchange_state"
+        ]
+        allowed_states = set(exchange_state_field.get_values(self.env))
+        for rec in self:
+            configured_states = rec._deduplicate_get_exchange_record_states()
+            invalid_states = sorted(configured_states - allowed_states)
+            if invalid_states:
+                raise ValidationError(
+                    self.env._(
+                        "Invalid exchange state(s): %(invalid_states)s. "
+                        "Allowed values are: %(allowed_states)s",
+                        invalid_states=", ".join(invalid_states),
+                        allowed_states=", ".join(sorted(allowed_states)),
+                    )
+                )
--- a/edi_exchange_deduplicate_oca/static/description/index.html
+++ b/edi_exchange_deduplicate_oca/static/description/index.html
@@ -411,6 +411,9 @@ the same record. If so, mark the oldest one as obsolete (except
 <ul class="simple">
 <li>“block_obsolescence” is an technical option on records to avoid
 marking them as obsolete.</li>
+<li>You can restrict deduplication to specific exchange states with the
+technical field “deduplicate_on_exchange_record_status”
+(comma-separated values from “edi_exchange_state”).</li>
 </ul>
 <p>With all the types that have been enabled “Delete obsolete records”
 option, the cron will remove their obsolete records.</p>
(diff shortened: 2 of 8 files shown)


edi_exchange_deduplicate_oca/models/edi_exchange_type.py
```diff
@@ -18,3 +19,32 @@ class EDIExchangeType(models.Model):
...
+        exchange_state_field = self.env["edi.exchange.record"]._fields[
+            "edi_exchange_state"
+        ]
+        allowed_states = set(exchange_state_field.get_values(self.env))
+        for rec in self:
+            configured_states = {
+                state.strip()
+                for state in (rec.deduplicate_on_exchange_record_status or "").split(",")
+                if state.strip()
+            }
```
simahawk:
```suggestion
            configured_states = rec._deduplicate_get_exchange_record_states()
            [....]
```
and move that bit to a reusable method to be used by `_edi_get_duplicates` too.

edi_exchange_deduplicate_oca/static/description/index.html
```diff
@@ -411,6 +411,9 @@ <h2><a class="toc-backref" href="#toc-entry-2">Usage</a></h2>
 <ul class="simple">
 <li>“block_obsolescence” is an technical option on records to avoid
 marking them as obsolete.</li>
+<li>You can restrict deduplication to specific exchange states with the
```
simahawk, line 414: next time `export SKIP=oca-gen-addon-readme` to avoid updating REAMEs

simahawk (commented):
LG overall

edi_exchange_deduplicate_oca/models/edi_exchange_type.py
wouitmil: Done

edi_exchange_deduplicate_oca/models/edi_exchange_type.py
```diff
@@ -18,3 +19,35 @@ class EDIExchangeType(models.Model):
         default=True,
         help="Delete records marked as obsolete.",
     )
+
+    deduplicate_on_exchange_record_status = fields.Char(
+        default="new,output_pending",
+        groups="base.group_no_one",
+    )
+
+    def _deduplicate_get_exchange_record_states(self):
```
Ricardoalso, line 28: A fallback on  ("new", "output_pending") for retro compatibility would be great

**EDIT**
probably in `_edi_get_duplicates` method

edi_exchange_deduplicate_oca/models/edi_exchange_type.py
```diff
@@ -18,3 +19,35 @@ class EDIExchangeType(models.Model):
         default=True,
         help="Delete records marked as obsolete.",
     )
+
+    deduplicate_on_exchange_record_status = fields.Char(
```
ivantodorovich, line 23: For UX, could we use a Many2many on `ir.model.fields.selection` + a domain to ensure it's limited to the `exchange.record.edi_exchange_state` field?

Kind of (but not really) like: https://github.com/odoo/odoo/blob/73e427150e097906046157395fc61c95a1729459/odoo/addons/base/models/ir_actions.py#L690-L691

Something like this (⚠️ untested and incomplete):

```py
deduplicate_on_exchange_record_status_ids = fields.Many2many(
    comodel_name="ir.model.fields.selection",
    domain=lambda self: [("field_id", "=", self.env["ir.model.fields"]._get("exchange.record", "edi_exchange_state").id)],
    ondelete="cascade",
    ...
)
```

As a many2many, you could use the `many2many_tags` widget on the view, which looks way nicer

wouitmil: Hello @ivantodorovich ,

It's something I proposed but @simahawk prefered to keep it simple.

edi_exchange_deduplicate_oca/models/edi_exchange_type.py
wouitmil: Hello @Ricardoalso , 

That's why I setted default="new,output_pending". Doesn't it seems sufficient for you?

edi_exchange_deduplicate_oca/models/edi_exchange_type.py
ivantodorovich: In matters of taste, there is no disputing

edi_exchange_deduplicate_oca/models/edi_exchange_type.py
Ricardoalso: Sorry, I didn’t see it earlier 🥲 

If `deduplicate_on_exchange_record_status` is empty (i.e., default values have been manually removed), we end up with the domain`("edi_exchange_state", "in", [])`, which matches zero records. As a result, the deduplication feature is effectively disabled by the state filter.

I am not against, but it might be worth documenting this behavior.

edi_exchange_deduplicate_oca/models/edi_exchange_type.py
simahawk: let's go how it is now

simahawk (approved):

edi_exchange_deduplicate_oca/models/edi_exchange_type.py
simahawk: I would add a safe fallback in this method

ivantodorovich (approved):

edi_exchange_deduplicate_oca/models/edi_exchange_type.py
Ricardoalso: Well, I guess it will be in another PR 👀

PR #306, opened 2026-06-11
[19.0][IMP] edi_core_oca: allow custom exchange action result

wouitmil:
Introduces a new class to wrap ``_exchange_$action`` methods. It stores the methods' results and a message allowing further customization.

### DIFF
--- a/edi_core_oca/models/edi_backend.py
+++ b/edi_core_oca/models/edi_backend.py
@@ -16,6 +16,7 @@ from odoo import exceptions, fields, models
 from odoo.exceptions import UserError
 
 from ..exceptions import EDINotImplementedError, EDIValidationError
+from ..utils import EdiExchangeActionResult
 
 _logger = logging.getLogger(__name__)
 
@@ -126,8 +127,12 @@ class EDIBackend(models.Model):
             # Remove file to regenerate
             exchange_record.exchange_file = False
         self._check_exchange_generate(exchange_record, force=force)
-        output = self._exchange_generate(exchange_record, **kw)
-        message = None
+        action_result = self._ensure_action_result(
+            self._exchange_generate(exchange_record, **kw),
+            default_message=exchange_record._exchange_status_message("generate_ok"),
+        )
+        output = action_result.output
+        message = action_result.message
         encoding = exchange_record.type_id.encoding or "UTF-8"
         encoding_error_handler = (
             exchange_record.type_id.encoding_out_error_handler or "strict"
@@ -142,7 +147,6 @@ class EDIBackend(models.Model):
                 }
             )
         if output:
-            message = exchange_record._exchange_status_message("generate_ok")
             try:
                 with self.env.cr.savepoint():
                     self._validate_data(exchange_record, output)
@@ -203,7 +207,16 @@ class EDIBackend(models.Model):
     def _exchange_generate(self, exchange_record, **kw):
         exchange_function = self._get_exec_handler(exchange_record, "generate")
         ctx = self._get_record_env_ctx(exchange_record, "generate")
-        return exchange_function(exchange_record.with_context(**ctx), **kw)
+        result = exchange_function(exchange_record.with_context(**ctx), **kw)
+        return self._ensure_action_result(
+            result,
+            default_message=exchange_record._exchange_status_message("generate_ok"),
+        )
+
+    def _ensure_action_result(self, result, default_message=None):
+        if isinstance(result, EdiExchangeActionResult):
+            return result
+        return EdiExchangeActionResult(output=result, message=default_message)
 
     # TODO: add tests
     def _validate_data(self, exchange_record, value=None, **kw):
@@ -244,9 +257,10 @@ class EDIBackend(models.Model):
         res = ""
         try:
             with self.env.cr.savepoint():
-                send_result = self._exchange_send(exchange_record)
+                send_result = self._ensure_action_result(
+                    self._exchange_send(exchange_record)
+                )
                 _logger.debug("%s sent", exchange_record.identifier)
-            _logger.debug("%s sent", exchange_record.identifier)
         except self._send_retryable_exceptions() as err:
             traceback = _get_exception_traceback()
             error = _get_exception_msg(err)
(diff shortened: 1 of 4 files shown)


edi_core_oca/models/edi_backend.py
```diff
@@ -529,7 +551,11 @@ def exchange_receive(self, exchange_record):
         message = None
         content = None
         try:
-            content = self._exchange_receive(exchange_record)
+            receive_result = self._ensure_action_result(
```
simahawk: It's used twice here. It should stay only in the private method.

Ricardoalso:
@wouitmil 
Rebased for the purposes of the origin project

wouitmil:
@ivantodorovich this PR is ready to be merged

ivantodorovich (approved):
Ok to you @simahawk ?

edi_core_oca/models/edi_backend.py
```diff
@@ -126,8 +127,12 @@ def exchange_generate(self, exchange_record, store=True, force=False, **kw):
             # Remove file to regenerate
             exchange_record.exchange_file = False
         self._check_exchange_generate(exchange_record, force=force)
-        output = self._exchange_generate(exchange_record, **kw)
-        message = None
+        action_result = self._ensure_action_result(
```
simahawk, line 130: each `_exchange_$action` should take care of returning the wrapped result.
Here you should not wrap.
Also I wouldn't pass `default_message`.
Just do `message = result.message or exchange_record._exchange_status_message("generate_ok")`

edi_core_oca/models/edi_backend.py
```diff
@@ -269,15 +284,18 @@ def exchange_send(self, exchange_record):
             res = "__sql_error__"
             raise
         else:
-            # TODO: maybe the send handler should return desired message and state
-            message = exchange_record._exchange_status_message("send_ok")
+            message = (
+                send_result.message
+                or send_result.output
```
simahawk, line 289: if you want the output as a message you must return it as the message

edi_core_oca/models/edi_backend.py
```diff
@@ -269,15 +284,18 @@ def exchange_send(self, exchange_record):
...
+                or exchange_record._exchange_status_message("send_ok")
+            )
             error = traceback = None
             state = (
                 "output_sent_and_processed"
                 if self.output_sent_processed_auto
                 else "output_sent"
             )
-            res = message
+            res = send_result.output or message
```
simahawk, line 298: read above

edi_core_oca/utils.py
```diff
@@ -3,6 +3,14 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
 import hashlib
+from dataclasses import dataclass
+from typing import Any
+
+
+@dataclass
+class EdiExchangeActionResult:
```
simahawk, line 11:
```suggestion
class EDIExchangeActionResult:
```

simahawk (changes requested):

ivantodorovich:
ping @wouitmil

SilvioC2C:
@simahawk @wouitmil @ivantodorovich I tried to implement Simone's suggestions, but there are a few issues with them, the main one being:

> each `_exchange_$action` should take care of returning the wrapped result

If we convert the `_exchange_$action` results to an `EDIExchangeActionResult` from within the `_exchange_$action` methods themselves or by using a decorator, most tests within this repo will fail because they either:
- expect the result to be a `str` and receive an `EDIExchangeActionResult` object
- or mock the method itself with `mocked_func.return_value = "something"`, which means that the public implementation `exchange_$action` will receive a `str` object instead of an `EDIExchangeActionResult` object, leading to a lot of `AttributeError` exceptions

Either way, tests fail, in both `edi_core_oca` and other modules that depend on it.

simahawk:
As per our discussion: nobody should override `_exchange_$action` methods as it would mess up w/ the handler mechanism. We should update relevant tests and move on w/ the decorator approach.

SilvioC2C:
@simahawk @wouitmil tests are fixed, should I squash all commits into the first one (except the last one, related to `edi_queue_oca`)?

edi_core_oca/utils.py
```diff
@@ -3,6 +3,46 @@
...
+
+    Such actions would usually return a simple ``str``; use this class to store the
+    result and add a custom message to it.
+    """
+
+    output: Any = None
+    message: str | None = None
+
+    @classmethod
+    def convert_result(cls, result: Any) -> "EDIExchangeActionResult":
```
simahawk: as it returns an instance of itself, I prefer
```suggestion
    def from_result(cls, result: Any) -> "EDIExchangeActionResult":
```

edi_core_oca/models/edi_backend.py
```diff
@@ -142,7 +145,8 @@ def exchange_generate(self, exchange_record, store=True, force=False, **kw):
                 }
             )
         if output:
-            message = exchange_record._exchange_status_message("generate_ok")
+            if not message:
```
simahawk: We already have a fallback on this above. Do we need this check?
BTW prior to this change the message was None till this point.
As the message from the handler should take precedence now, I'd say we can drop this line.

edi_core_oca/utils.py
```diff
@@ -3,6 +3,46 @@
...
+        """Wraps ``result```` in an ``EDIExchangeActionResult``
+
+        By default, instances created using this method will contain no message.
+        """
+        return result if isinstance(result, cls) else cls(output=result, message=None)
+
+    @classmethod
+    def wrap_result(cls):
```
simahawk: As there's no param to pass, do we need it to be a callable?

simahawk (commented):
@SilvioC2C yes pls and rewrite them as

```
[IMP] edi_core_oca: allow custom action result

...motivation...
```
and

```
[IMP] edi_queue_oca: adapt to new action result
```

simahawk (approved):
LGTM, @etobella I'd like to get your opinion on this change

etobella (approved):
For me it seems right. Thanks!

simahawk:
/ocabot merge minor

@SilvioC2C can you backport to 18 pls?

SilvioC2C:
@simahawk backported in #346

PR #308, opened 2026-06-15
[FIX] edi_core_oca: fix tests

etobella:
An error happens due to this change.

https://github.com/odoo/odoo/commit/6fd3703fde9d81da69cb6fad999644204f517a56

Now, `:` is transform into `-`


LoisRForgeFlow (approved):
👍

SilvioC2C (approved):

PR #309, opened 2026-06-15
[FIX] Initialize response variable in edi_backend.py

imlopes:


imlopes:
replaced by: https://github.com/OCA/edi-framework/pull/310

PR #312, opened 2026-06-19
[19.0][IMP] edi_queue_oca: add timezone-aware daily ETA scheduling

Ricardoalso:
This pull request introduces an enhancement to the EDI Queue OCA module by allowing EDI exchange jobs to be scheduled and accumulated for release at a specific daily time, in addition to existing per-type channel and priority configuration. 

**New job scheduling and accumulation features:**

* Added fields to `edi.exchange.type` for enabling ETA scheduling, configuring execution hour, minute, and timezone, and logic to compute the next scheduled runtime in UTC. This allows all jobs for a given exchange type to be held and released together at a fixed time each day, supporting off-peak processing and trading partner requirements.
* Modified job dispatch logic so that, if ETA scheduling is enabled, jobs are queued with the calculated ETA; otherwise, jobs are dispatched immediately as before.

**Accumulating jobs until a fixed daily time**

Enabling ETA scheduling on an exchange type causes every job created for that type to be held in the queue until the configured time of day, rather than being processed immediately. All jobs that arrive during the day accumulate and are released together at that scheduled time.

Typical use cases:

- A trading partner's receiving system processes incoming files only during a specific nightly window (e.g., 22:00).
- Resource-intensive EDI operations (large exports, heavy transformations) are deferred to off-peak hours to avoid competing with daytime workloads.
- An operational preference to send a batch of documents at a predictable daily time instead of dispatching them one by one in real time.

**The execution time is configured using three fields:**
- Hour — hour of the day (00–23)
- Minute — minute of the hour (00–59)
- Timezone — the timezone in which the hour and minute are interpreted; defaults to the current user's timezone

At runtime, the configured time is converted to the next matching UTC datetime and set as the queue job ETA. If the target time for the current day has already passed, the job is automatically scheduled for the same time on the next day.

**When ETA scheduling is disabled, jobs are dispatched immediately as usual.**

### DIFF
--- a/edi_queue_oca/models/edi_exchange_record.py
+++ b/edi_queue_oca/models/edi_exchange_record.py
@@ -47,6 +47,8 @@ class EdiExchangeRecord(models.Model):
         priority = exchange_type.job_priority
         if priority:
             params["priority"] = priority
+        if exchange_type.job_eta_enabled and (eta := exchange_type._get_job_eta()):
+            params["eta"] = eta
         # Avoid generating the same job for the same record if existing
         params["identity_key"] = exchange_record_job_identity_exact
         return params
--- a/edi_queue_oca/models/edi_exchange_type.py
+++ b/edi_queue_oca/models/edi_exchange_type.py
@@ -2,7 +2,14 @@
 # Copyright 2025 Dixmit
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
 
-from odoo import fields, models
+from datetime import datetime
+
+import pytz
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, fields, models
+
+from odoo.addons.base.models.res_partner import _tz_get as timezone_selection
 
 
 class EdiExchangeType(models.Model):
@@ -12,3 +19,64 @@ class EdiExchangeType(models.Model):
         comodel_name="queue.job.channel",
     )
     job_priority = fields.Integer()
+    job_eta_enabled = fields.Boolean(
+        string="Enable ETA Scheduling",
+        help="Accumulate all queue jobs for this exchange type and release them "
+        "at the daily time configured below, instead of dispatching each job "
+        "immediately. Use this when a trading partner only processes files at a "
+        "fixed daily window, or to concentrate resource-intensive EDI work in "
+        "off-peak hours.",
+    )
+    job_eta_hour = fields.Selection(
+        [(str(x).zfill(2), str(x).zfill(2)) for x in range(24)],
+        string="Execution time - hour",
+        help="Hour of the day at which jobs for this type should be scheduled.",
+        default="00",
+    )
+    job_eta_minute = fields.Selection(
+        [(str(x).zfill(2), str(x).zfill(2)) for x in range(60)],
+        string="Execution time - minute",
+        help="Minute of the hour at which jobs for this type should be scheduled.",
+        default="00",
+    )
+    job_eta_tz = fields.Selection(
+        timezone_selection,
+        string="Execution time - timezone",
+        help="ETA's timezone for jobs of this type",
+        default=lambda self: self._get_default_eta_tz(),
+    )
+
+    @api.model
+    def _get_default_eta_tz(self):
+        tz_list = timezone_selection(self)
+        return self.env.user.tz or (tz_list and tz_list[0][0]) or "UTC"
+
+    def _get_job_eta(self):
+        """Returns the ETA for the current type as timezone-naive datetime"""
+        self.ensure_one()
+        if (
+            not self.job_eta_enabled
+            or not self.job_eta_tz
+            or not self.job_eta_hour
+            or not self.job_eta_minute
+        ):
+            return None
+
+        self.ensure_one()
+        now = pytz.UTC.localize(datetime.utcnow())
+        eta = (
+            pytz.timezone(self.job_eta_tz)
+            .localize(
+                datetime(
+                    now.year,
+                    now.month,
+                    now.day,
+                    int(self.job_eta_hour),
+                    int(self.job_eta_minute),
+                )
+            )
+            .astimezone(pytz.UTC)
+        )
+        if eta < now:
+            eta += relativedelta(days=1)
+        return eta.replace(tzinfo=None)
--- a/edi_queue_oca/tests/test_exchange_type.py
+++ b/edi_queue_oca/tests/test_exchange_type.py
@@ -1,0 +1,60 @@
+# Copyright 2026 Camptocamp SA
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from datetime import datetime
+
+from freezegun import freeze_time
+
+from .common import EDIQueueCommonTestCase
+
+
+class EDIExchangeTypeTestCase(EDIQueueCommonTestCase):
+    def setUp(self):
+        super().setUp()
+        self.record = self._make_record()
+
+    def _get_job_eta(self, record):
+        return self._get_delayed(record).delayable.eta
+
+    def test_job_eta_disabled_no_eta_applied(self):
+        """job_eta_enabled=False: no ETA is set regardless of value."""
+        self.exchange_type_in.job_eta_hour = "22"
+        self.exchange_type_in.job_eta_minute = "00"
+        # job_eta_enabled defaults to False
+        self.assertIsNone(self._get_job_eta(self.record))
+
+    @freeze_time("2024-01-15 20:00:00")
+    def test_job_eta_scheduled_same_day(self):
+        """ETA not yet reached today: job lands on the same calendar day."""
+        self.env.user.tz = "UTC"
+        self.exchange_type_in.job_eta_enabled = True
+        self.exchange_type_in.job_eta_hour = "22"
+        self.exchange_type_in.job_eta_minute = "00"
+        self.assertEqual(
+            self._get_job_eta(self.record),
+            datetime(2024, 1, 15, 22, 0, 0),
+        )
+
+    @freeze_time("2024-01-15 23:00:00")
+    def test_job_eta_scheduled_next_day(self):
+        """ETA already passed today: job rolls over to the next calendar day."""
+        self.env.user.tz = "UTC"
+        self.exchange_type_in.job_eta_enabled = True
+        self.exchange_type_in.job_eta_hour = "22"
+        self.exchange_type_in.job_eta_minute = "00"
+        self.assertEqual(
+            self._get_job_eta(self.record),
+            datetime(2024, 1, 16, 22, 0, 0),
+        )
+
+    @freeze_time("2024-01-15 20:00:00")
+    def test_job_eta_midnight_schedules_next_day(self):
+        """Midnight (00:00) schedules for the next 00:00 in user TZ."""
+        self.env.user.tz = "UTC"
+        self.exchange_type_in.job_eta_enabled = True
+        self.exchange_type_in.job_eta_hour = "00"
+        self.exchange_type_in.job_eta_minute = "00"
+        self.assertEqual(
+            self._get_job_eta(record=self.record),
+            datetime(2024, 1, 16, 0, 0, 0),
+        )
--- a/edi_queue_oca/tests/test_record.py
+++ b/edi_queue_oca/tests/test_record.py
@@ -3,18 +3,18 @@
 # @author: Simone Orsi <simahawk@gmail.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
+from datetime import datetime
 
-from odoo.addons.edi_core_oca.tests.common import EDIBackendCommonTestCase
+import pytz
+
 from odoo.addons.queue_job.delay import DelayableRecordset
 
+from .common import EDIQueueCommonTestCase
 
-class EDIRecordTestCase(EDIBackendCommonTestCase):
+
+class EDIRecordTestCase(EDIQueueCommonTestCase):
     def test_with_delay_override(self):
-        vals = {
-            "model": self.partner._name,
-            "res_id": self.partner.id,
-        }
-        record = self.backend.create_record("test_csv_input", vals)
+        record = self._make_record()
         parent_channel = self.env["queue.job.channel"].create(
             {
                 "name": "parent_test_chan",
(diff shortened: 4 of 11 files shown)


edi_queue_oca/models/edi_exchange_record.py
```diff
@@ -38,6 +41,28 @@ def _register_hook(self):
...
+        """Convert a decimal-hour ETA to the next occurrence as a naive UTC datetime.
+
+        ``eta_time`` is a float in [0, 24) where the fractional part represents
+        minutes (e.g. 22.5 → 22:30). The target is expressed in the current
+        user's timezone and rolled forward by one day when the time has already
+        passed today.
+
+        Returns a naive UTC datetime suitable for ``queue.job`` ``eta`` param.
+        """
+        hours, minutes = divmod(round(eta_time * 60), 60)
```
vvrossem: If `eta_time = 23.992`, then `round(23.992 * 60) = round(1439.52) = 1440`, giving `hours = 24`. `datetime.replace(hour=24)` raises` ValueError`. The constraint allows any value in `[0, 24), but values above `~23.9916` will round up to 1440 minutes = 24:00 and crash at runtime. The fix is to cap or handle the rollover:

```suggestion
        total_minutes = round(eta_time * 60) % 1440
        hours, minutes = divmod(total_minutes, 60)
```

edi_queue_oca/models/edi_exchange_type.py
```diff
@@ -12,3 +13,24 @@ class EdiExchangeType(models.Model):
         comodel_name="queue.job.channel",
     )
     job_priority = fields.Integer()
+    eta_time = fields.Float(
+        string="Execute at (timezone aware)",
+        help="Hour of the day (decimal) at which jobs for this type should be "
+        "scheduled, in your local timezone. Use decimal fractions for minutes "
+        "(e.g. 22.5 = 22:30). Must be in the range [0, 24). "
+        "Leave empty (0) to disable.",
+    )
```
vvrossem: use `-1.0` to disable? or a boolean "enabled" flag ?

vvrossem: can the user set 00:00 midnight as value?

edi_queue_oca/models/edi_exchange_type.py
```diff
@@ -12,3 +13,24 @@ class EdiExchangeType(models.Model):
...
+
+    @api.constrains("eta_time")
+    def _check_eta_time(self):
+        for rec in self:
+            if rec.eta_time and not (0 <= rec.eta_time < 24):
+                raise ValidationError(
+                    self.env._(
+                        "'%(name)s': Execute at time %(value).2f is not a valid "
+                        "24-hour value. Enter a decimal hour in the range [0, 24) "
```
vvrossem: nitpicking:
```suggestion
                        "24-hour value. Enter a decimal hour in the range [0, 24[ "
```

edi_queue_oca/models/edi_exchange_record.py
```diff
@@ -38,6 +41,28 @@ def _register_hook(self):
     def action_exchange_send_job_options(self):
         return {"priority": 0}
 
+    def _eta_float_to_utc(self, eta_time):
```
simahawk: move it to `utils.py` and should have some unit tests

edi_queue_oca/models/edi_exchange_type.py
```diff
@@ -12,3 +13,24 @@ class EdiExchangeType(models.Model):
         comodel_name="queue.job.channel",
     )
     job_priority = fields.Integer()
+    eta_time = fields.Float(
+        string="Execute at (timezone aware)",
+        help="Hour of the day (decimal) at which jobs for this type should be "
```
simahawk: This means that for a certain type you want jobs to accumulate till a certain hour, correct?
What is the use-case? It would be nice to mention that in the readme as well.

simahawk (changes requested):
In addition to the existing remarks

edi_queue_oca/models/edi_exchange_record.py
Ricardoalso: Nice edge catch

edi_queue_oca/models/edi_exchange_type.py
Ricardoalso: I went for the flag solution and yes the users can set 00:00 midnight as value

edi_queue_oca/models/edi_exchange_type.py
Ricardoalso: Exactly, my specific business use-case is to accumulate resource-intensive EDI work into off-peak hours. 
I pushed a fixup commit to document it

vvrossem (approved):

edi_queue_oca/models/edi_exchange_type.py
```diff
@@ -12,3 +13,31 @@ class EdiExchangeType(models.Model):
...
     job_priority = fields.Integer()
+    eta_enabled = fields.Boolean(
+        string="Enable ETA Scheduling",
+        help="Accumulate all queue jobs for this exchange type and release them "
+        "at the daily time configured below, instead of dispatching each job "
+        "immediately. Use this when a trading partner only processes files at a "
+        "fixed daily window, or to concentrate resource-intensive EDI work in "
+        "off-peak hours.",
+    )
+    eta_time = fields.Float(
```
SilvioC2C: Personally, I'd split this field into 2 different fields, one for hours (values from 0 to 23) and another for minutes (values from 0 to 59), and add a field `eta_tz` that defines the timezone; then, simply have a computed, non-stored field to compute the ETA:
```
from datetime import datetime
from dateutil.relativedelta import relativedelta
from pytz import timezone, utc

from odoo import api, fields, models
from odoo.addons.base.models.res_partner import _tzs as timezone_selection


class EdiExchangeType(models.Model):
    [...]

    eta_hour = fields.Selection(
        list((x, str(x).zfill(2)) for x in range(24))),
        string="Execution time - hour",
        help="Hour of the day at which jobs for this type should be scheduled.",
        required=True,
        default="00",
    )
    eta_minute = fields.Selection(
        list((x, str(x).zfill(2)) for x in range(60))),
        string="Execution time - minute",
        help="Minute of the hour at which jobs for this type should be scheduled.",
        required=True,
        default="00",
    )
    eta_tz = fields.Selection(
        timezone_selection,
        string="Execution time - timezone",
        help="ETA's timezone for jobs of this type",
        required=True,
        default=lambda self: self._get_default_eta_tz(),
    )

    @api.model
    def _get_default_eta_tz(self):
        return self.env.user.tz or (timezone_selection and timezone_selection[0]) or "UTC"

    def _get_eta(self):
        """Returns the ETA for the current type as timezone-naive datetime"""
        self.ensure_one()
        now = utc.localize(datetime.now())
        eta = timezone(self.eta_time_tz).localize(datetime(now.year, now.month, now.day, self.eta_time_hour, self.eta_time_minute)).astimezone(utc)
        if eta < now:
            eta += relativedelta(days=1)
        return eta.replace(tzinfo=None)
```

SilvioC2C (commented):
The idea is good, but I would go with a cleaner approach:
- use separate fields for hours and minutes instead of parsing strings
- let the user define the ETA TZ
- don't add utils, let the record itself compute its own ETA

edi_queue_oca/models/edi_exchange_type.py
```diff
@@ -12,3 +13,31 @@ class EdiExchangeType(models.Model):
         comodel_name="queue.job.channel",
     )
     job_priority = fields.Integer()
+    eta_enabled = fields.Boolean(
```
simahawk: pls prefix all job fields w/ `job_*`

edi_queue_oca/models/edi_exchange_type.py
```diff
@@ -12,3 +13,31 @@ class EdiExchangeType(models.Model):
...
+        string="Execute at (timezone aware)",
+        help="Hour of the day (decimal) at which jobs for this type should be "
+        "scheduled, in your local timezone. Use decimal fractions for minutes "
+        "(e.g. 22.5 = 22:30, 0.0 = 00:00 midnight). Must be in the range [0, 24[.",
+    )
+
+    @api.constrains("eta_enabled", "eta_time")
+    def _check_eta_time(self):
+        for rec in self:
+            if rec.eta_enabled and not (0 <= rec.eta_time < 24):
```
grindtildeath: Use `float_compare` ?

edi_queue_oca/utils.py
```diff
@@ -14,3 +19,28 @@ def exchange_record_job_identity_exact(job_):
...
+    minutes (e.g. 22.5 → 22:30, 0.0 → 00:00 midnight). The target is expressed
+    in the current user's timezone and rolled forward by one day when the time
+    has already passed today.
+
+    Returns a naive UTC datetime suitable for ``queue.job`` ``eta`` param.
+    """
+    # Use timedelta from midnight rather than datetime.replace(hour=...) so that
+    # values near 24.0 (e.g. 23.992 → 1440 total minutes) wrap cleanly to 00:00
+    # the next day instead of crashing with ValueError: hour must be in 0..23.
+    total_minutes = round(eta_time * 60) % 1440
```
grindtildeath: IMO it would be cleaner to build a datetime.time object that you can use to combine with datetime object. cf https://github.com/OCA/server-tools/blob/19.0/base_time_window/models/time_window_mixin.py#L125-L128 (to convert the float_time to time object)

grindtildeath (commented):
Small things

edi_queue_oca/models/edi_exchange_type.py
Ricardoalso: Thanks for the suggestions! I initially took the more user-friendly approach using float_time, which is also used across core Odoo modules.
But I like the simplicity of having both `eta_hour` and `eta_minute` fields. Since this is linked to a technical setting, I think we can trade user-friendliness for a simpler and more efficient design.

edi_queue_oca/utils.py
Ricardoalso: Thanks for pointing this code out ! :rocket: 
I finally decided to take the approach suggested by Silvio https://github.com/OCA/edi-framework/pull/312#discussion_r3458353930

grindtildeath (approved):
Looks much better 👍

edi_queue_oca/models/edi_exchange_type.py
```diff
@@ -12,3 +19,62 @@ class EdiExchangeType(models.Model):
...
+        string="Execution time - hour",
+        help="Hour of the day at which jobs for this type should be scheduled.",
+        required=True,
+        default="00",
+    )
+    job_eta_minute = fields.Selection(
+        [(str(x).zfill(2), str(x).zfill(2)) for x in range(60)],
+        string="Execution time - minute",
+        help="Minute of the hour at which jobs for this type should be scheduled.",
+        required=True,
```
simahawk: ⚠️ none of these fields should be required at field level.
Add form modifiers and validators if needed.

edi_queue_oca/tests/test_record.py
```diff
@@ -26,12 +41,70 @@ def test_with_delay_override(self):
         )
         self.exchange_type_in.job_channel_id = channel
         self.exchange_type_in.job_priority = 5
-        # re-enable job delayed feature
-        delayed = record.with_context(queue_job__no_delay=False).with_delay()
-        # Silent useless warning
-        # `Delayable Delayable(edi.exchange.record*) was prepared but never delayed`
-        delayed.delayable._generated_job = object()
+        self.exchange_type_in.job_eta_enabled = True
```
simahawk, line 49: I wouldn't touch the original test. A new one should be added to test your case.

edi_queue_oca/tests/test_record.py
```diff
@@ -26,12 +41,70 @@ def test_with_delay_override(self):
...
+            hour=22, minute=0, second=0, microsecond=0
+        )
+        expected_eta = target_22h_user.astimezone(utc_tz).replace(tzinfo=None)
+        self.assertEqual(job_eta, expected_eta)
         self.assertTrue(isinstance(delayed, DelayableRecordset))
         self.assertEqual(delayed.recordset, record)
         self.assertEqual(delayed.delayable.channel, "root.parent_test_chan.test_chan")
         self.assertEqual(delayed.delayable.priority, 5)
+
+    def test_eta_disabled_no_eta_applied(self):
```
simahawk: here you are mixing unit tests for the eta functions and fields w/ the jobs integration.
This is not a test for "no_eta_applied".
I would move this stuff to `test_exchange_type.py` and do this test as the original one (well, w/o touching it as mentioned above).

simahawk (changes requested):

edi_queue_oca/tests/test_record.py
Ricardoalso: Fix-up commit to refactor the tests and reorganize the code. The main logic of `test_with_delay_override` untouched, I just consolidated reusable blocks.

edi_queue_oca/tests/test_record.py
Ricardoalso: New file` edi_queue_oca/tests/test_exchange_type.py` created :heavy_check_mark:

edi_queue_oca/tests/test_exchange_type.py
```diff
@@ -0,0 +1,60 @@
...
+class EDIExchangeTypeTestCase(EDIQueueCommonTestCase):
+    def _get_job_eta(self, record):
+        return self._get_delayed(record).delayable.eta
+
+    def test_job_eta_disabled_no_eta_applied(self):
+        """job_eta_enabled=False: no ETA is set regardless of value."""
+        self.exchange_type_in.job_eta_hour = "22"
+        self.exchange_type_in.job_eta_minute = "00"
+        # job_eta_enabled defaults to False
+        record = self._make_record()
```
simahawk: nitpicking: the record do not have any impact on this feature, it seems useless to create one record all the times. It could be done at class setup.

simahawk (approved):

simahawk:
squash pls

edi_queue_oca/tests/test_exchange_type.py
Ricardoalso: Indeed, I took the opportunity to make a small improvement while squashing :wink:

ivantodorovich:
@simahawk this should be ready to merge

PR #317, opened 2026-07-07
[17.0][ADD]edi_notification_related_record_oca: configurable per-action related-record notifications

GuillemCForgeFlow:
Makes the chatter notes `edi_oca` posts on an exchange record's related records
configurable per exchange type and per action.

**Why:** every exchange logs a note on the business record when its data is generated and
again when it is sent — two near-redundant entries for a non-technical reader, and the ones
that matter end up buried. They stay useful while setting a flow up, so rather than
dropping them, which events are surfaced is left to configuration.

#### What

1. **`[IMP] edi_oca`** — `_notify_related_record` now receives the action the note comes
   from (`generate`/`send`/`process`/`receive`, or `None` for acknowledgements). It carries
   no behaviour in core; it is there so an override can decide per action.
2. **`[ADD] edi_notification_related_record_oca`** — adds
   `notify_related_record_on_{generate,send,process,receive}` on `edi.exchange.type`,
   gated in a single override, with a *Related Record Notifications* page added to the
   form via `inherit_id` (no existing element is moved). All four default to `True`, so
   install changes nothing and no migration is needed. Unticking one suppresses the note
   only: the record still advances, keeps its state and errors, and still triggers events.

**On the `process` toggle:** deliberately not filtered by direction. `exchange_process` is
input-only, but an output record's "processed" note comes from the `check` component
resolved in `edi_backend._exchange_output_check_state` (e.g. `edi_storage_oca` polling a
remote `done`/`error` file), which also calls `_notify_done`/`_notify_error`.
Acknowledgements are not bound to an action and have no toggle (`readme/ROADMAP.md`).

**Tests:** the default note, each of generate/send/process disabled, that disabling one
action leaves the others posting, and that an action with no toggle still notifies.

cc @ForgeFlow

### DIFF
--- a/edi_oca/models/edi_exchange_type.py
+++ b/edi_oca/models/edi_exchange_type.py
@@ -185,6 +185,26 @@ class EDIExchangeType(models.Model):
         "(default is always 'Raise Error').",
     )
     allow_empty_files_on_receive = fields.Boolean(string="Allow Empty Files")
+    notify_related_record_on_generate = fields.Boolean(
+        string="Notify related record on generate",
+        default=True,
+        help="Post a note on the related record when the exchange is generated.",
+    )
+    notify_related_record_on_send = fields.Boolean(
+        string="Notify related record on send",
+        default=True,
+        help="Post a note on the related record when the exchange is sent.",
+    )
+    notify_related_record_on_process = fields.Boolean(
+        string="Notify related record on process",
+        default=True,
+        help="Post a note on the related record when the exchange is processed.",
+    )
+    notify_related_record_on_receive = fields.Boolean(
+        string="Notify related record on receive",
+        default=True,
+        help="Post a note on the related record when the exchange is received.",
+    )
 
     _sql_constraints = [
         (
@@ -285,6 +305,18 @@ class EDIExchangeType(models.Model):
         if hasattr(exchange_record.record, "_get_edi_exchange_record_name"):
             return exchange_record.record._get_edi_exchange_record_name(exchange_record)
         return slugify(exchange_record.record.display_name)
+
+    def _notify_related_record_on(self, action):
+        """Whether a chatter note should be posted on the related record.
+
+        Gated per stage via the `notify_related_record_on_*` toggles. Actions without a
+        dedicated toggle (e.g. `ack`) keep notifying to preserve legacy behavior.
+        """
+        self.ensure_one()
+        field_name = "notify_related_record_on_%s" % action
+        if field_name not in self._fields:
+            return True
+        return self[field_name]
 
     def is_partner_enabled(self, partner):
         """Check if given partner record is allowed for the current type.
--- a/edi_oca/views/edi_exchange_type_views.xml
+++ b/edi_oca/views/edi_exchange_type_views.xml
@@ -73,6 +73,27 @@
                                         widget="many2many_tags"
                                     />
                                 </group>
+                                <group
+                                    name="notifications"
+                                    string="Related-record notifications"
+                                >
+                                    <field
+                                        name="notify_related_record_on_generate"
+                                        invisible="direction == 'input'"
+                                    />
+                                    <field
+                                        name="notify_related_record_on_send"
+                                        invisible="direction == 'input'"
+                                    />
+                                    <field
+                                        name="notify_related_record_on_process"
+                                        invisible="direction == 'output'"
+                                    />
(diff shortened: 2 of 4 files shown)


edi_oca/models/edi_exchange_type.py
```diff
@@ -185,6 +185,26 @@ class EDIExchangeType(models.Model):
         "(default is always 'Raise Error').",
     )
     allow_empty_files_on_receive = fields.Boolean(string="Allow Empty Files")
+    notify_rel_rec_on_generate = fields.Boolean(
```
SilvioC2C: Nitpicking: improve naming
```suggestion
    notify_related_record_on_generate = fields.Boolean(
```

edi_oca/models/edi_exchange_type.py
```diff
@@ -286,6 +306,18 @@ def _get_record_name(self, exchange_record):
             return exchange_record.record._get_edi_exchange_record_name(exchange_record)
         return slugify(exchange_record.record.display_name)
 
+    def _notify_related_on(self, action):
```
SilvioC2C: Nitpicking: improve naming
```
    def _notify_related_record_on(self, action):
```

edi_oca/views/edi_exchange_type_views.xml
```diff
@@ -39,32 +39,66 @@
...
-                            />
-                            <field
-                                name="encoding_in_error_handler"
-                                invisible="direction == 'output'"
-                            />
-                            <field
-                                name="allow_empty_files_on_receive"
-                                invisible="direction == 'output'"
-                            />
-                        </group>
```
SilvioC2C, line 65: Please don't do this: this architecture might be inherited and customized by third-party modules, and this change might break the view's inheritance mechanism (even though I personally agree a `page` in the `notebook` yields to a better UX).

edi_oca/models/edi_exchange_type.py
GuillemCForgeFlow: changes applied 👍🏿

edi_oca/models/edi_exchange_type.py
GuillemCForgeFlow: I agree this is more readable 👌🏿

edi_oca/views/edi_exchange_type_views.xml
GuillemCForgeFlow: I see, thanks for pointing out. what I've done now is still keep the `name="config"` attribute on the main group in the view. I understand there's still a chance that will can break current installations. do you think we should perhaps note this to be done in the next migration?

GuillemCForgeFlow:
@simahawk any thoughts on this?

simahawk (changes requested):
@GuillemCForgeFlow Thanks for your contrib!
I don't think this should stay core.
I would add a new module or add this to  `edi_notification_oca`.

GuillemCForgeFlow:
> @GuillemCForgeFlow Thanks for your contrib! I don't think this should stay core. I would add a new module or add this to `edi_notification_oca`.

@simahawk I've created a new module to handle the initial proposal, it lives now under a new one and not `edi_notification_oca` as I've seen those features unrelated to my changes. Please feel free to share your thoughts on this 🙂

simahawk:
@GuillemCForgeFlow thanks for your proposal.

AFAIU your final goal is to - for instance - avoid having a message "Exchange record generated" followed right away by a "Exchange record sent". Correct?

I'd like to summarize the use cases that we have here and in the current implementation of edi_notification.

UC

1. be able to notify for failure certain groups and users (already in edi_notif)
2. be able to turn on/off message posting for specific actions (what we have here)

Do you see any other use case?

GuillemCForgeFlow:
> @GuillemCForgeFlow thanks for your proposal.
> 
> AFAIU your final goal is to - for instance - avoid having a message "Exchange record generated" followed right away by a "Exchange record sent". Correct?
> 
> I'd like to summarize the use cases that we have here and in the current implementation of edi_notification.
> 
> UC
> 
> 1. be able to notify for failure certain groups and users (already in edi_notif)
> 2. be able to turn on/off message posting for specific actions (what we have here)
> 
> Do you see any other use case?

No, exactly as you described. the final goal is what drove me to have this new module implemented so that we can let the user manage the settings and have more granularity on what to show.

do you think that makes sense?

PR #319, opened 2026-07-13
[19.0[IMP] edi_core_oca: mark process integrity errors as failed

yankinmax:


wouitmil (approved):

edi_core_oca/models/edi_backend.py
Ricardoalso, line 1: Some time ago I added the line `except (OperationalError, IntegrityError):` https://github.com/OCA/edi-framework/blob/19.0/edi_core_oca/models/edi_backend.py#L266 PR https://github.com/OCA/edi-framework/pull/240 in order to fix failed transactions where the cursor enters an aborted state.

With this new commit we will never reach the conditional structure `if res != "__sql_error__": `https://github.com/OCA/edi-framework/blob/19.0/edi_core_oca/models/edi_backend.py#L282

simahawk (changes requested):
I agree w/ @Ricardoalso

edi_core_oca/models/edi_backend.py
yankinmax: Agree with you, thanks for the review

edi_core_oca/models/edi_backend.py
```diff
@@ -486,7 +493,14 @@ def exchange_process(self, exchange_record):
             error = _get_exception_msg(err)
             state = "input_processed_error"
             res = f"Error: {error}"
-        except (OperationalError, IntegrityError):
+        except IntegrityError as err:
```
Ricardoalso, line 496: Well with this change we just end up duplicating code

```
if self.env.context.get("_edi_process_break_on_error"):
    raise
traceback = _get_exception_traceback()
error = _get_exception_msg(err)
state = "input_processed_error"
res = f"Error: {error}"
```
I rechecked the code and with the changes introduced by 8f444874, 
I would keep IntegrityError inside `_swallable_exceptions` because it always runs inside a `cr.savepoint()`, which rolls back to the savepoint before this is caught, leaving the cursor usable.

Leaving OperationalError handling (`res = "__sql_error__"`) separately as it may indicate a broken connection rather than a "bad statement".

@guewen your input is welcome on this one 🙏

simahawk:
What is the status here? :)

Ricardoalso:
> What is the status here? :)

This is being handled by https://github.com/OCA/queue/pull/955 and https://github.com/OCA/edi-framework/pull/324

This PR can be closed

yankinmax:
> > What is the status here? :)
> 
> This is being handled by [OCA/queue#955](https://github.com/OCA/queue/pull/955) and #324
> 
> This PR can be closed

Thanks for the info, I'll close this one!

PR #321, opened 2026-07-13
[19.0][FIX] edi_queue_oca: propagate failed jobs to exchanges

yankinmax:
When an EDI queue job reached its maximum retries, only the queue job was marked as failed while the related exchange remained pending.

Propagate terminal send, receive, and process job failures to the corresponding exchange error state, including the exception message and traceback.

### DIFF
--- a/edi_queue_oca/models/queue_job.py
+++ b/edi_queue_oca/models/queue_job.py
@@ -1,0 +1,28 @@
+# Copyright 2026 Camptocamp SA
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+
+from odoo import models
+
+
+class QueueJob(models.Model):
+    _inherit = "queue.job"
+
+    def write(self, vals):
+        result = super().write(vals)
+        if vals.get("state") == "failed":
+            self._mark_related_edi_exchanges_failed()
+        return result
+
+    def _mark_related_edi_exchanges_failed(self):
+        """Propagate terminal EDI job failures to their exchange records."""
+        supported_methods = {
+            "action_exchange_process",
+            "action_exchange_receive",
+            "action_exchange_send",
+        }
+        jobs = self.filtered(
+            lambda job: job.model_name == "edi.exchange.record"
+            and job.method_name in supported_methods
+        )
+        for job in jobs:
+            job.records.sudo()._mark_failed_from_queue_job(job)
--- a/edi_queue_oca/models/__init__.py
+++ b/edi_queue_oca/models/__init__.py
@@ -1,3 +1,4 @@
 from . import edi_exchange_record
 from . import edi_exchange_type
 from . import edi_backend
+from . import queue_job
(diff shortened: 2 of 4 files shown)


yankinmax:
Hello @simahawk can you pls take a look?

SilvioC2C (approved):

edi_queue_oca/models/queue_job.py
```diff
@@ -0,0 +1,28 @@
...
+
+from odoo import models
+
+
+class QueueJob(models.Model):
+    _inherit = "queue.job"
+
+    def write(self, vals):
+        result = super().write(vals)
+        if vals.get("state") == "failed":
```
Ricardoalso, line 12: Maybe there’s an opportunity to create an "on-failure" hook in the queue_job module

https://github.com/OCA/queue/blob/19.0/queue_job/models/queue_job.py#L268

Ricardoalso, line 12: Direct SQL updates executed by the queue_job runner will bypass this logic https://github.com/OCA/queue/blob/19.0/queue_job/jobrunner/runner.py#L217


 In the situation where _the number of retries exceeds the maximum allowed, the job is marked as "failed" with the error "JobFoundDead."_

grindtildeath: FWIW when the job is failing, only the job object will be updated through a dedicated function for failure,
and then the modified values are going to be stored on the odoo record: https://github.com/OCA/queue/blob/19.0/queue_job/controllers/main.py#L187-L188

edi_queue_oca/models/queue_job.py
```diff
@@ -0,0 +1,28 @@
...
+            self._mark_related_edi_exchanges_failed()
+        return result
+
+    def _mark_related_edi_exchanges_failed(self):
+        """Propagate terminal EDI job failures to their exchange records."""
+        supported_methods = {
+            "action_exchange_process",
+            "action_exchange_receive",
+            "action_exchange_send",
+        }
```
grindtildeath, line 22: Why don't we have `action_exchange_generate` ❓ 

I'd suggest to have a constant variable with the functions list, so that it can be relied on here and in the `_register_hook`.

grindtildeath (commented):
Idea looks good, but IMO there's still something wrong because:

- We have this https://github.com/OCA/edi-framework/blob/18.0/edi_core_oca/models/edi_backend.py#L247-L253 in the core module although we do not depend on queue_job.
- What we try to solve here is to have the exchange record in an error state when the number of retries from a job is exhausted.

grindtildeath:
Here's another attempt at solving the same issue: https://github.com/OCA/edi-framework/pull/323

simahawk (changes requested):
not a viable solution

edi_queue_oca/models/queue_job.py
grindtildeath: On fail hooks in queue job: https://github.com/OCA/queue/pull/955

grindtildeath:
Alternative implementation:  #324

simahawk:
closing in favor of https://github.com/OCA/edi-framework/pull/324

PR #327, opened 2026-08-13
[FIX] edi_record_metadata_oca: hide metadata page when empty

ivantodorovich:
Hide the Metadata page when the record has no metadata, to avoid showing this:

Screenshot 2026-08-13 at 2 07 03 PM


simahawk (approved):
Thanks! Can you backport it pls?

PR #329, opened 2026-08-18
[18.0] Party data improvements

simahawk:
See atomic commits for details.

Impact: only for the ones using party data in exchange templates. There the display_name will be replace w/ name but that's not a problem for 99% of the cases.

Spoiler: once this is merged I'll drop component dependency for basic party data and deprecate `edi_exchange_template_party_data`.


imlopes (approved):
LGTM

santostelmo (approved):

PR #330, opened 2026-08-24
[18.0] Add edi_party_helper_oca, deprecate edi_party_data_oca

simahawk:
Extract EDI party-data lookup out of the `component` framework so it can be used standalone.

- New module `edi_party_helper_oca`: `EDIParty`, a plain dataclass (no `component` dependency) that computes party data (`name`, `identifiers`, `endpoint`, `lang`) and behaves like a read-only dict for backward compat. Exposed via a new `edi.party.helper` abstract model (`get_party(exchange_record, partner, **kw)`). Moved the `id_category_ids` field/view here too.
- `edi_party_data_oca`: `edi.party.data.mixin` component now just delegates to `edi.party.helper`, kept only for backward compatibility.
- Both `edi_party_data_oca` and `edi_exchange_template_party_data` marked deprecated: readme notes + a `post_init_hook` logging a deprecation message on install.
- Tests split accordingly: exhaustive data tests moved to `edi_party_helper_oca`, component tests trimmed to wiring-only checks.


simahawk:
test fails only on license check

edi_exchange_template_party_data/hooks.py
```diff
@@ -0,0 +1,13 @@
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+def post_init_hook(env):
+    _logger.info(
```
SilvioC2C, line 7: Nitpicking: `_logger.warning()` would be better, imho

edi_exchange_template_party_data/__manifest__.py
```diff
@@ -12,4 +12,5 @@
     "website": "https://github.com/OCA/edi-framework",
     "depends": ["edi_exchange_template_oca", "edi_party_data_oca"],
     "auto_install": True,
+    "post_init_hook": "post_init_hook",
```
SilvioC2C: Two things:
- since the module is getting deprecated, shouldn't it make sense to remove the `auto_install` flag?
- `post_init_hook` is only executed upon module installation: would it make sense to add a `post_load` hook too to display the deprecation warning whenever the module is loaded?

edi_party_data_oca/__manifest__.py
```diff
@@ -12,8 +12,6 @@
     "website": "https://github.com/OCA/edi-framework",
     "author": "Camptocamp, Odoo Community Association (OCA)",
     "maintainers": ["simahawk"],
-    "depends": ["edi_component_oca", "partner_identification"],
-    "data": [
-        "views/edi_exchange_type_views.xml",
-    ],
+    "depends": ["edi_component_oca", "edi_party_helper_oca"],
+    "post_init_hook": "post_init_hook",
```
SilvioC2C: Same as before: maybe add a `post_load` script too?

edi_party_data_oca/hooks.py
```diff
@@ -0,0 +1,13 @@
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+def post_init_hook(env):
+    _logger.info(
```
SilvioC2C, line 7: Same as above: `_logger.warning()` might be a better choice

SilvioC2C (approved):
Overall LGTM, a couple of minor remarks

edi_exchange_template_party_data/hooks.py
simahawk: done on purpose to not bother builds w/ warnings (which would cause failure)

edi_exchange_template_party_data/__manifest__.py
simahawk: very good points. In fact I wanted to use post_load but I forgot about it :sweat: 
And for the auto_install flag: correct! I'll fix them thanks!

edi_exchange_template_party_data/hooks.py
SilvioC2C: Thanks, you can ignore my comment then (I'm trying to resolve it but GitHub shows an error msg instead)

edi_exchange_template_party_data/__manifest__.py
simahawk: fixed, thanks!

PR #333, opened 2026-08-31
[19.0] [IMP] edi_core_oca: scheduled edi.configuration triggers

ivantodorovich:
Add `on_cron_hourly` and `on_cron_daily` triggers, fired by scheduled actions. They carry no originating record, so each configuration resolves its own targets: a global one runs once, any other runs against the records linked to it through a many2many on its model.

### DIFF
--- a/edi_core_oca/demo/edi_configuration_demo.xml
+++ b/edi_core_oca/demo/edi_configuration_demo.xml
@@ -1,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="edi_conf_trigger_cron_hourly" model="edi.configuration.trigger">
+        <field name="name">On hourly schedule</field>
+        <field name="code">on_cron_hourly</field>
+        <field
+            name="description"
+        >Trigger when the hourly EDI scheduled action runs</field>
+    </record>
+    <record id="edi_conf_trigger_cron_daily" model="edi.configuration.trigger">
+        <field name="name">On daily schedule</field>
+        <field name="code">on_cron_daily</field>
+        <field
+            name="description"
+        >Trigger when the daily EDI scheduled action runs</field>
+    </record>
+
+    <record id="cron_edi_configuration_hourly" model="ir.cron">
+        <field name="name">EDI configuration hourly trigger</field>
+        <field name="active" eval="True" />
+        <field name="user_id" ref="base.user_root" />
+        <field name="interval_number">1</field>
+        <field name="interval_type">hours</field>
+        <field
+            name="nextcall"
+            eval="(DateTime.now() + timedelta(hours=1)).strftime('%Y-%m-%d %H:00:00')"
+        />
+        <field name="model_id" ref="edi_core_oca.model_edi_configuration" />
+        <field name="state">code</field>
+        <field name="code">model._cron_run_by_trigger("on_cron_hourly")</field>
+    </record>
+    <record id="cron_edi_configuration_daily" model="ir.cron">
+        <field name="name">EDI configuration daily trigger</field>
+        <field name="active" eval="True" />
+        <field name="user_id" ref="base.user_root" />
+        <field name="interval_number">1</field>
+        <field name="interval_type">days</field>
+        <field
+            name="nextcall"
+            eval="(DateTime.now() + timedelta(days=1)).strftime('%Y-%m-%d 00:00:00')"
+        />
+        <field name="model_id" ref="edi_core_oca.model_edi_configuration" />
+        <field name="state">code</field>
+        <field name="code">model._cron_run_by_trigger("on_cron_daily")</field>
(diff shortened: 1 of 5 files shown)


edi_core_oca/tests/test_edi_configuration.py
```diff
@@ -86,6 +88,152 @@ def _setup_records(cls):  # pylint:disable=missing-return
...
+                "snippet_do": (
+                    'env["edi.exchange.consumer.test"].create('
+                    '{"name": "By global conf"})'
+                ),
+            }
+        )
+
+        self.edi_configuration._cron_run_by_trigger("on_cron_hourly")
+
+        self.assertEqual(
```
simahawk: I would use `RecordCapturer`. Not a blocker.

simahawk (changes requested):
Nice addition. Very handy to export batches or reports with a given cadence.
Can you please document it in CONFIGURE.md?

Questions:
1. no idempotency: if someone runs the cron again, the configs will be executed again -> wouldn't be worth to mention it in an info box?
2. despite there's no dependency addition I wonder if we want to host this feature in core. In particular I wonder if we shouldn't simply supply the cron method and leave implementers the freedom to create specific triggers and crons. All in all what makes it work is the trigger code, no matter what it is.

ivantodorovich:
Thanks @simahawk !

> Can you please document it in CONFIGURE.md?

Sure thing, I'll update asap

> 1. no idempotency: if someone runs the cron again, the configs will be executed again -> wouldn't be worth to mention it in an info box?

An info box in the edi.configuration page?
The problem is how to identify scheduled action triggers in order to display/hide this box. A naive solution would be by name (startswith `on_cron`), but it's not technically correct.. people have the freedom to implement their own scheduled action triggers and use whatever name they want

> 2. despite there's no dependency addition I wonder if we want to host this feature in core. In particular I wonder if we shouldn't simply supply the cron method and leave implementers the freedom to create specific triggers and crons. All in all what makes it work is the trigger code, no matter what it is.

Yes that's the idea, implementers have the freedom to create new triggers and crons, but I figured we can already supply an hourly and daily cron triggers to serve as implementation example whilst covering the most common usage

ivantodorovich:
it's ready

ivantodorovich:
@simahawk changed again: crons and triggers are demo data only, not shipped directly

simahawk (approved):

edi_core_oca/demo/edi_configuration_demo.xml
SilvioC2C, line 1: These records are declared in `/demo` because they're supposed to be imported when creating a DB with demo data, am I right?
If not, and we're declaring them for tests only, I'd rather have them created by the test's `setUpClass()` method instead.
Non-blocking.

SilvioC2C (approved):
One minor doubt, else LGTM!

edi_core_oca/demo/edi_configuration_demo.xml
ivantodorovich: That's correct, this is actual demo data

Ricardoalso (approved):

edi_core_oca/models/edi_configuration.py
```diff
@@ -239,22 +242,56 @@ def edi_get_conf_global(self, exchange_record, trigger):
...
+    def _get_records(self, model):
+        """Return the records of ``model`` subscribed to this configuration.
+
+        A record subscribes to a configuration through a many2many declared on
+        its own model -- there may be several such relations, one per EDI flow,
+        and a model holding none yields no record.
+        """
+        self.ensure_one()
+        return model.search(
+            Domain.OR(
```
simahawk, line 282: @ivantodorovich getting back to this in the scope of backporting to v18: do we really need to use `Domain`?

ivantodorovich: I wouldn't do any different in 19.0+, as `Domain` is the best idiomatic way to define them

But no, in previous versions it can be expressions.OR with list-defined domains

simahawk: My point is that this domain is very simple and I doubt there's anything to be optimized and we could make this compatible w/ both versions and ease back/fwd porting.
We could have:

```
try:
    from odoo.fields import Domain
    OR = Domain.OR
except ImportError:
    from odoo.osv import expression
    OR = expression.OR
[...]
OR([fname, "in", self.ids])
```
WDYT?

ivantodorovich: I'd say version-specific code lives in the version-specific branch
I wouldn't be afraid of a backport/forwardport that requires a code change

Ricardoalso: I can backport it with the specific code if needed :smile:

ivantodorovich:
@simahawk can we merge this? 🙏🏻

PR #334, opened 2026-09-02
[19.0] [IMP] edi_core_oca, edi_queue_oca: chain receive and process

ivantodorovich:
An input record was received and then left for the next run of the input cron to be processed, up to an hour later. Chain the two the way generate and send already are, so a document is processed as soon as it lands.

### DIFF
--- a/edi_queue_oca/models/edi_exchange_record.py
+++ b/edi_queue_oca/models/edi_exchange_record.py
@@ -103,6 +103,13 @@ class EdiExchangeRecord(models.Model):
         job1.on_done(self.delayable(priority=0).action_exchange_send())
         job1.delay()
 
+    def action_exchange_receive_process_chained(self):
+        job1 = self.delayable().action_exchange_receive()
+        # Chain process job.
+        # Raise prio to max to process the record as soon as it lands.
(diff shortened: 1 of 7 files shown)


ivantodorovich:
@simahawk is this what you had in mind?

edi_queue_oca/models/edi_exchange_record.py
```diff
@@ -103,6 +103,13 @@ def action_exchange_generate_send_chained(self):
         job1.on_done(self.delayable(priority=0).action_exchange_send())
         job1.delay()
 
+    def action_exchange_receive_process_chained(self):
```
simahawk, line 106: don't we have a test on chained jobs for the other case?

simahawk (commented):
LG overall

edi_queue_oca/models/edi_exchange_record.py
ivantodorovich: I added more tests

simahawk (approved):

simahawk:
Thanks! Can you please backport to v18? :pray: 
/ocabot merge minor

PR #336, opened 2026-09-03
[19.0] [FIX] edi_queue_oca: on fail hook with non string exception message

ivantodorovich:
The hook joins `exc_name` and `exc_message` as strings, but the message stored by queue_job is the first argument of the exception, which is for example the errno for an OSError, and it may also be missing.

### DIFF
--- a/edi_queue_oca/models/edi_exchange_record.py
+++ b/edi_queue_oca/models/edi_exchange_record.py
@@ -117,10 +117,15 @@ class EdiExchangeRecord(models.Model):
 
     def _job_on_fail_update(self, failed_state, **kw):
         self.ensure_one()
+        # ``exc_message`` is the first argument of the exception: it is not
+        # always a string (e.g. the errno of an ``OSError``) and may be missing.
+        error = ": ".join(
+            str(kw[k]) for k in ("exc_name", "exc_message") if kw.get(k) is not None
+        )
         self.write(
             {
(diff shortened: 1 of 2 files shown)


ivantodorovich:
ping @simahawk @grindtildeath

grindtildeath (approved):

edi_queue_oca/models/edi_exchange_record.py
```diff
@@ -117,10 +117,15 @@ def _job_on_fail_process(self, **kw):
 
     def _job_on_fail_update(self, failed_state, **kw):
         self.ensure_one()
+        # ``exc_message`` is the first argument of the exception: it is not
+        # always a string (e.g. the errno of an ``OSError``) and may be missing.
+        error = kw.get("exc_name") or ""
+        if kw.get("exc_message") is not None:
+            error = f"{error}: {kw['exc_message']}"
```
simahawk: You can have `: FooErr`.
```suggestion
        error = []
        for k in ("exc_name", "exc_message"):
            if kw.get(k):
                error.append(kw[k])
        error = ": ".join(error)
```
or put and IF inside the f-string but is not really readable.

ivantodorovich: done

simahawk (approved):

PR #203, opened 2025-09-22
[17.0][IMP] `edi_endpoint_oca`: querystring params usage

SilvioC2C:
Compatibility w/ https://github.com/OCA/web-api/pull/114

PR #337, opened 2026-09-07
[IMP] edi_endpoint_oca: querystring params usage

yankinmax:
Compatibility w/ https://github.com/OCA/web-api/pull/114

This PR supersedes (just a cherry-pick to have it opened and get merged):
- [ ] https://github.com/OCA/edi-framework/pull/203

### DIFF
--- a/edi_endpoint_oca/models/edi_endpoint.py
+++ b/edi_endpoint_oca/models/edi_endpoint.py
@@ -67,9 +67,9 @@ class EDIEndpoint(models.Model):
                     _("Exchange type not compatible with selected backend type.")
                 )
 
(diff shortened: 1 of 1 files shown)


SilvioC2C:
@yankinmax can you cherry-pick the rest of the commits too?
Or, as an alternative, check @simahawk suggestion in #203?

PS: after a discussion w/ Simone, we think it'd be better to proceed without this feature, as it's most probably unnecessary, and it will become just more code to maintain in the long term.
If already in use by some custom project, please check/test first whether Simone's suggestions [here](https://github.com/OCA/web-api/pull/114) can be implemented instead.

PR #340, opened 2026-09-08
[19.0][IMP] edi_queue_oca: cancel wait_dependencies jobs on exchange failure

AaronHForgeFlow:
When a chained queue job (such as generate->send or receive->process) fails, downstream dependent jobs remain permanently stuck in wait_dependencies because queue_job only cancels dependent jobs upon explicit job cancellation.

Ensure _job_on_fail_update cancels any wait_dependencies queue jobs belonging to the errored exchange record so zombie jobs do not linger.

### DIFF
--- a/edi_queue_oca/README.rst
+++ b/edi_queue_oca/README.rst
@@ -64,22 +64,22 @@ Per-type job configuration
 
 Each **Exchange Type** gains a *Queue* tab with optional settings:
 
-+---------------------------+------------------------------------------+
-| Field                     | Purpose                                  |
-+===========================+==========================================+
-| **Job channel**           | Route jobs to a specific channel (e.g.   |
-|                           | ``root.edi.high``).                      |
-+---------------------------+------------------------------------------+
-| **Job priority**          | Integer priority passed to the queue job |
-|                           | (lower = higher priority).               |
-+---------------------------+------------------------------------------+
-| **Enable ETA Scheduling** | Toggle to activate daily job             |
-|                           | accumulation (see below).                |
-+---------------------------+------------------------------------------+
-| **Execution time**        | Hour, minute, and timezone at which      |
-|                           | accumulated jobs are released. Visible   |
-|                           | only when ETA Scheduling is enabled.     |
-+---------------------------+------------------------------------------+
++---------------------------+-----------------------------------------+
+| Field                     | Purpose                                 |
++===========================+=========================================+
+| **Job channel**           | Route jobs to a specific channel (e.g.  |
+|                           | ``root.edi.high``).                     |
++---------------------------+-----------------------------------------+
+| **Job priority**          | Integer priority passed to the queue    |
+|                           | job (lower = higher priority).          |
++---------------------------+-----------------------------------------+
+| **Enable ETA Scheduling** | Toggle to activate daily job            |
+|                           | accumulation (see below).               |
++---------------------------+-----------------------------------------+
+| **Execution time**        | Hour, minute, and timezone at which     |
+|                           | accumulated jobs are released. Visible  |
+|                           | only when ETA Scheduling is enabled.    |
++---------------------------+-----------------------------------------+
 
 Accumulating jobs until a fixed daily time
 ------------------------------------------
@@ -91,20 +91,20 @@ during the day accumulate and are released together at that moment.
 
 **Typical use cases:**
 
-- A trading partner's receiving system only processes incoming files at
-  a specific nightly window (e.g. 22:00).
-- Resource-intensive EDI operations (large exports, heavy
-  transformations) should be deferred to off-peak hours to avoid
-  competing with daytime workloads.
-- Operational preference to send a batch of documents at a predictable
-  daily time instead of dispatching them one by one in real time.
+-  A trading partner's receiving system only processes incoming files at
+   a specific nightly window (e.g. 22:00).
+-  Resource-intensive EDI operations (large exports, heavy
+   transformations) should be deferred to off-peak hours to avoid
+   competing with daytime workloads.
+-  Operational preference to send a batch of documents at a predictable
+   daily time instead of dispatching them one by one in real time.
 
 The execution time is configured with three fields:
 
-- **Hour** — hour of the day (00–23).
-- **Minute** — minute of the hour (00–59).
-- **Timezone** — the timezone in which the hour and minute are
-  interpreted. Defaults to the current user's timezone.
+-  **Hour** — hour of the day (00–23).
+-  **Minute** — minute of the hour (00–59).
+-  **Timezone** — the timezone in which the hour and minute are
+   interpreted. Defaults to the current user's timezone.
 
 At runtime the configured time is converted to the next matching UTC
 datetime and set as the queue job ETA. If the target time for today has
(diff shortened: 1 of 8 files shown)


simahawk:
If I get it right, when a send job fails you will cancel it... I don't think this is ideal at all.
If you cancel it, you'll miss it and you won't be able to re-schedule it.
Am I missing anything?

AaronHForgeFlow:
This targets only downstream child jobs in state wait_dependencies (for example, the chained send job when generate has failed) When generate fails, the child send job is stuck in wait_dependencies waiting for its parent generate to reach done

should I mark the job as failed instead?

simahawk:
I see this useful only in the situation where a second pair of jobs does the work and leaves the old ones stale:

* job11: generate -> fails
* job12: send -> stale
* job21: generate -> done
* job22: send -> done

If I get it right, with your change as soon as a generate fails, the send will be discarded leaving only the generate job in failed state. If you run only this job, no send will happen unless you also rely on crons but that can cause additional delays. Correct?


> should I mark the job as failed instead?

only if requeueing the generate job causes the requeueing of the chained job too (which I doubt).

AaronHForgeFlow:
> If you run only this job, no send will happen unless you also rely on crons but that can cause additional delays. Correct?

Yes, correct. In my case I usually rely on the user action, they will create a new exchange record leaving the previous one as canceled or failed. I am open to suggestions, I am not sure if this is the best approach.

simahawk:
TBH I don't have a precise answer. Options:
1. when a full process is over input/output can spawn a GC job to cancel stale jobs for that record
2. add GC to the existing crons to cancel stale jobs
3. any other option?

I tend to say that opt 2 is the cleanest and safest.

AaronHForgeFlow:
Added a system parameter so the zombie jobs can be collected some time. If this is too aggresive maybe we can add a configuration or a new field in the backend so it does not affect current instances.

edi_queue_oca/models/edi_backend.py
```diff
@@ -20,3 +27,66 @@ def _send_retryable_exceptions(self):
...
+        res = super()._cron_check_output_exchange_sync(**kw)
+        self._gc_stale_exchange_jobs()
+        return res
+
+    def _cron_check_input_exchange_sync(self, **kw):
+        res = super()._cron_check_input_exchange_sync(**kw)
+        self._gc_stale_exchange_jobs()
+        return res
+
+    def _gc_stale_exchange_jobs_grace_hours(self):
```
simahawk: can you pls prefix all method names w/ `_job_*`?

edi_queue_oca/models/edi_backend.py
```diff
@@ -20,3 +27,66 @@ def _send_retryable_exceptions(self):
...
         return RetryableJobError
+
+    def _cron_check_output_exchange_sync(self, **kw):
+        res = super()._cron_check_output_exchange_sync(**kw)
+        self._gc_stale_exchange_jobs()
+        return res
+
+    def _cron_check_input_exchange_sync(self, **kw):
+        res = super()._cron_check_input_exchange_sync(**kw)
+        self._gc_stale_exchange_jobs()
```
simahawk: This means that the GC will run twice and we don't have any way to turn it off if needed.
Options:
1. move it to its own cron that can be tuned/disabled if needed (disabled by default)
2. if the config param is zero the whole feature is disable (default)

edi_queue_oca/models/edi_backend.py
```diff
@@ -20,3 +27,66 @@ def _send_retryable_exceptions(self):
...
+        # will ever wake this job up again.
+        blocking = [
+            parent_states.get(uuid)
+            for uuid in parent_uuids
+            if parent_states.get(uuid) != "done"
+        ]
+        return bool(blocking) and not any(x in LIVE_JOB_STATES for x in blocking)
+
+    def _gc_stale_exchange_jobs(self):
+        """Cancel exchange jobs that can never be woken up.
```
simahawk, line 76: I like to respect this PEP for multi-line docstrings https://peps.python.org/pep-0257/#multi-line-docstrings

edi_queue_oca/models/edi_backend.py
```diff
@@ -20,3 +27,66 @@ def _send_retryable_exceptions(self):
...
+        parent_states = {
+            job.uuid: job.state
+            for job in job_model.search([("uuid", "in", list(parent_uuids))])
+        }
+        stale = candidates.filtered(
+            lambda job: self._is_stale_exchange_job(job, parent_states)
+        )
+        if stale:
+            _logger.info("EDI exchange GC: cancelling %d stale jobs.", len(stale))
+            stale.button_cancelled()
```
simahawk: shall we post a message or override the `result` message to make clear that it was canceled by GC?

simahawk (changes requested):

edi_queue_oca/models/edi_backend.py
AaronHForgeFlow: good point, chose option 1

edi_queue_oca/README.rst
```diff
@@ -143,11 +155,11 @@ Authors
...
-- Simone Orsi <simahawk@gmail.com>
-- Enric Tobella <enric.tobella@dixmit.com>
-- Manuel Regidor <manuel.regidor@sygel.es>
-- Thien Vo <thienvh@trobz.com>
-- Jordi Masvidal <jordi.masvidal@forgeflow.com>
+-  Simone Orsi <simahawk@gmail.com>
+-  Enric Tobella <enric.tobella@dixmit.com>
+-  Manuel Regidor <manuel.regidor@sygel.es>
+-  Thien Vo <thienvh@trobz.com>
+-  Jordi Masvidal <jordi.masvidal@forgeflow.com>
```
simahawk, line 162: @AaronHForgeFlow time to add yourself? :)

simahawk (approved):
LGTM

SilvioC2C (approved):

edi_queue_oca/models/edi_backend.py
```diff
@@ -20,3 +35,67 @@ def _send_retryable_exceptions(self):
...
+        param = self.env["ir.config_parameter"].sudo()
+        return int(param.get_param("edi_queue_oca.gc_stale_jobs_grace_hours", 24))
+
+    def _job_gc_stale_domain(self, grace_hours):
+        deadline = fields.Datetime.now() - timedelta(hours=grace_hours)
+        return [
+            ("model_name", "=", self.exchange_record_model._name),
+            ("state", "=", WAIT_DEPENDENCIES),
+            ("date_created", "<=", deadline),
+        ]
```
Ricardoalso:
```suggestion
        return Domain([
            ("model_name", "=", self.exchange_record_model._name),
            ("state", "=", WAIT_DEPENDENCIES),
            ("date_created", "<=", deadline),
        ])
```

Ricardoalso (approved):

edi_queue_oca/models/edi_backend.py
simahawk: no please, it will make backporting harder for no real improvement (it's a simple domain, but correct me if I'm wrong)

Ricardoalso: It's not a blocking recommendation but, 
It is the recommended approach in the version 19.0 (security, safety, and readability advantages) and it improves the readability and usage of bitwise operations if this method needs to get overriden

AaronHForgeFlow: I did the change. But it is true it makes backporting harder. @simahawk I can revert the change, just let me know.

PR #341, opened 2026-09-09
[19.0] [IMP] edi_exchange_deduplicate_oca: deduplicate on reception

ivantodorovich:
Rename "Deduplicate on Send" to "Deduplicate on Exchange" and apply it to input types too.
The old field stays as a related alias, with a migration.

⚠️ Module version bumped manually for the migration script. Use nobump for merging

### DIFF
--- a/edi_exchange_deduplicate_oca/models/edi_exchange_type.py
+++ b/edi_exchange_deduplicate_oca/models/edi_exchange_type.py
@@ -1,18 +1,34 @@
 # Copyright 2024 Camptocamp
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+import logging
+
 from odoo import api, fields, models
 from odoo.exceptions import ValidationError
+
+_logger = logging.getLogger(__name__)
 
 
 class EDIExchangeType(models.Model):
     _inherit = "edi.exchange.type"
 
+    deduplicate_on_exchange = fields.Boolean(
+        string="Deduplicate on Exchange",
+        default=False,
+        help="When an exchange record is created, mark the older ones for the "
+        "same record that are still pending as obsolete, so only the freshest "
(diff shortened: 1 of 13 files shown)


edi_exchange_deduplicate_oca/models/edi_exchange_type.py
```diff
@@ -8,11 +8,22 @@
...
+        "the latest one matters.",
+    )
+    # Kept for backwards compatibility.
+    # TODO: Remove this alias in 20.0 migration
     deduplicate_on_send = fields.Boolean(
         string="Deduplicate on Send",
-        default=False,
-        help="Before sending an exchange record, check if a fresher one does not "
-        "exist for same record; if so, mark oldest one as obsolete.",
+        related="deduplicate_on_exchange",
```
simahawk: Do we really need this as a field? What are we trying to make compatible? Custom code?
If yes, I'd rather drop the field and add a property that logs deprecation.
If you keep the field, I would make it computed instead and log deprecation.

simahawk (commented):
LG overall. Minor remark.

edi_exchange_deduplicate_oca/models/edi_exchange_type.py
ivantodorovich: There's existing custom code out there defining their exchange types in data/ xml files, referencing the old `deduplicate_on_send` field. I don't want to break that.

I can add a deprecation warning, sure :)

ivantodorovich: it's ready ✅

simahawk (approved):

simahawk:
/ocabot merge nobump

@ivantodorovich could you backport to 18 pls?

PR #342, opened 2026-09-10
[FIX] [19.0] edi_core_oca: fix missing kw in super for _search(...)

acsonefho:
**Issue:**
Missing `**kw` given to the `super` call.

**How to reproduce:**
Generate a edi.exchange.record with a parent and archive this parent.

When you try to open the form view of the child, you will have an `Access Error` because Odoo try to load the parent but as all args are not given to the `super`, the parent is not returned and Odoo perform this as a non-accessible record.

`Fix:`
Just give the `**kw` during call to `super`.

In this case the `kw = {'active_test': False}`

This change comes from Odoo who give this as a parameter in v`19` (it was in the `context` in previous version):
`19`: https://github.com/odoo/odoo/blob/19.0/odoo/orm/models.py#L3802
`18`: https://github.com/odoo/odoo/blob/18.0/odoo/models.py#L4155

### DIFF
--- a/edi_core_oca/models/edi_exchange_record.py
+++ b/edi_core_oca/models/edi_exchange_record.py
@@ -576,10 +576,7 @@ class EDIExchangeRecord(models.Model):
 
     def _search(self, domain, offset=0, limit=None, order=None, **kw) -> Query:
         query = super()._search(
-            domain=domain,
-            offset=offset,
-            limit=limit,
-            order=order,
+            domain=domain, offset=offset, limit=limit, order=order, **kw
         )
         if self.env.is_superuser():
             # restrictions do not apply for the superuser


etobella (approved):

simahawk (approved):

simahawk:
Thanks!

acsonefho:
@etobella @simahawk thank you for the quick review and merge :rocket: 

I don't know if these 2 lines should be fixed too:
https://github.com/OCA/edi-framework/blob/19.0/edi_core_oca/models/edi_exchange_record.py#L634
https://github.com/OCA/edi-framework/blob/19.0/edi_core_oca/models/edi_exchange_record.py#L646

simahawk:
> @etobella @simahawk thank you for the quick review and merge 🚀
> 
> I don't know if these 2 lines should be fixed too: https://github.com/OCA/edi-framework/blob/19.0/edi_core_oca/models/edi_exchange_record.py#L634 https://github.com/OCA/edi-framework/blob/19.0/edi_core_oca/models/edi_exchange_record.py#L646



To have full ctx

case 2
```
            allowed = list(
                self.env[model]
                .with_context(active_test=False)
                ._search([("id", "in", recs.ids)])
            )
``` 
case 3
```
        if len(orig_ids) == limit and len(result) < len(orig_ids):
            extend_query = self._search(
                domain,
                offset=offset + len(orig_ids),
                limit=limit,
                order=order,
            )
``` 
For case 2 if passing active_test through context is still working we can keep it that way but might be worth to keep it consistent.
For case 3 we only miss `bypass_access` that in this context we don't care I think.

@etobella you might have a better picture on this part. I'll let you judge.

PR #314, opened 2026-06-29
[19.0][MIG] edi_gs1_oca

acsonefho:
Based on work done here: https://github.com/OCA/edi/pull/411

Depends on: https://github.com/OCA/edi-framework/pull/343

Probably a lot of improvements to do but the purpose of this PR is only the migration.

PR #343, opened 2026-09-10
[MIG][19.0] edi_xml_oca (from 18.0)

acsonefho:
Migration of `edi_xml_oca`

Improvement (into dedicated commits)
- Use `file_open` from Odoo instead of `open(...)`;
- Add `_listify(...)` to force having a `list` (when element have only a single value (but could have more; depending on `xsd`) `xmltodict` parse it like a `dict` instead of `list` of `dict`.


acsonefho:
@simahawk @etobella Can you review this one please? :pray:

edi_xml_oca/models/edi_xml.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def parse_xml(self, file_content, **kw):
+        """Read XML content.
+        :param file_content: str of XML file
+        :return: dict with final data
+        """
+        return self._xml_string_to_dict(file_content, **kw)
+
+    @staticmethod
+    def _listify(value):
```
simahawk, line 48: I'm not sure to understand what is the purpose of this method. 
Where is the issue coming from?
And as I pointed out on the GS1 PR we do not use `xmltodict` anymore but maybe we missed the fwd port of this PR to 17 and 18 https://github.com/OCA/edi-framework/pull/129

edi_xml_oca/models/edi_xml.py
```diff
@@ -0,0 +1,87 @@
...
+        iterating on such an element need a stable shape.
+
+        :param value: value read from a parsed XML dict
+        :return: list, empty when the element is missing
+        """
+        if value is None:
+            return []
+        return value if isinstance(value, list) else [value]
+
+    def validate(self, schema_path, xml_content, raise_on_fail=False):
```
simahawk, line 62: This handler should inherit from `edi.oca.handler.input.validate"` and `edi.oca.handler.output.validate` and implement the respective methods by respecting the signature.
That can only accept an exchange record. Everything else, like the schema, should come from configuration.
Quick solution: pass it over context and you can inject it as it is today via advanced settings with `env_ctx`.
Also. `_check_with_xsd` already supports lookup via ir.attachment.

simahawk (changes requested):

edi_xml_oca/models/edi_xml.py
acsonefho: I think there's a confusion: #129 replaced `xmlschema` with `lxml` for the XSD validation, but it actually introduced `xmltodict` for the parsing (`lxml` doesn't product `dict`). It has been forward-ported to 18.0 already, and this migration is based on it. What you asked me to remove in #314 was `xmlschema`, which is done.

About `_listify`: `xmlschema.to_dict()` was schema-aware, so an element declared with `maxOccurs > 1` always came back as a `list`. `xmltodict` doesn't know the schema: if the document holds a single occurence, you get a `dict` instead of a `list` of `dict` and the caller iterating on it breaks. `_listify` normalizes that in one place.

PR #344, opened 2026-09-10
[19.0][FIX] edi_core_oca: allow viewing archived exchange records

AaronHForgeFlow:
Avoids ORM AccessError when a specific record ID is requested in an action domain or form view without active_test=False

Odoo also returns archived records in several actions as well:

https://github.com/odoo/odoo/blob/18.0/addons/payment/models/payment_provider.py#L442-L451
https://github.com/odoo/odoo/blob/18.0/addons/crm/models/crm_lost_reason.py#L25-L33

### DIFF
--- a/edi_core_oca/models/edi_exchange_consumer_mixin.py
+++ b/edi_core_oca/models/edi_exchange_consumer_mixin.py
@@ -259,7 +259,8 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
         self.ensure_one()
         xmlid = "edi_core_oca.act_open_edi_exchange_record_view"
         action = self.env["ir.actions.act_window"]._for_xml_id(xmlid)
-        action["domain"] = [("id", "in", self.exchange_record_ids.ids)]
+        records = self.with_context(active_test=False).exchange_record_ids
+        action["domain"] = [("id", "in", records.ids)]
         # Purge default search filters from ctx to avoid hiding records
         ctx = action.get("context", {})
         if isinstance(ctx, str):
@@ -267,6 +268,7 @@ class EDIExchangeConsumerMixin(models.AbstractModel):
         action["context"] = {
             k: v for k, v in ctx.items() if not k.startswith("search_default_")
         }
+        action["context"]["active_test"] = False
         # Drop ID otherwise the context will be loaded from the action's record :S
         action.pop("id")
         return action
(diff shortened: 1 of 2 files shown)


edi_core_oca/models/edi_exchange_consumer_mixin.py
```diff
@@ -259,14 +259,16 @@ def action_view_edi_records(self):
         self.ensure_one()
         xmlid = "edi_core_oca.act_open_edi_exchange_record_view"
         action = self.env["ir.actions.act_window"]._for_xml_id(xmlid)
-        action["domain"] = [("id", "in", self.exchange_record_ids.ids)]
+        records = self.with_context(active_test=False).exchange_record_ids
```
simahawk, line 262: I would add a comment that mention the archival of records view backend conf.

Thinking of that, a few questions:
* is the counter already counting archived records? I guess so because they are related records
* are records fully visible in the related records tab?
* shall we syml-link the "active" flat on related record? Maybe just for display purposes?

AaronHForgeFlow: 1. Yes. exchange_record_count in EDIExchangeConsumerMixin runs _read_group on edi.exchange.related.record. Since edi.exchange.related.record has no active field, it counts all related records regardless of the target exchange record is archived.
2. On the related record, for example a purchase order, there is an EDI stat button.
3. If we add a stored active field to edi.exchange.related.record Odoo would hide the related records when an exchange is archived. Given that backends can automatically archive old exchange records via auto_archive_records_after_day, I prefer to avoid this from happening.

Ricardoalso (approved):

edi_core_oca/tests/test_consumer_mixin.py
```diff
@@ -107,6 +106,10 @@ def test_mixin(self):
                 exchange_record.type_id, self.backend
             )
         )
+        self.assertEqual(action["context"].get("active_test"), False)
+        exchange_record.active = False
+        action = self.consumer_record.action_view_edi_records()
+        self.assertIn(exchange_record.id, action["domain"][0][2])
```
SilvioC2C, line 112: Can you please add a non-regression tests?
- read field `self.consumer_record.exchange_record_count`
- archive the `exchange_record`
- invalidate the cached values of `self.consumer_record.exchange_record_[ids|count]`
- read again field `self.consumer_record.exchange_record_count`: the count should not change

AaronHForgeFlow: added :)

PR #349, opened 2026-09-18
[16.0][IMP] edi_oca: optimize exchange file search domains

santostelmo:
Searching on exchange_file (Binary, attachment=True) generates SQL subqueries on ir_attachment (IN/NOT IN on res_id), which is costly in batch cron flows.

This change introduces a stored computed boolean has_exchange_file on edi.exchange.record, computed from attachment metadata (res_model, res_field, res_id) via read_group, and switches backend domains to use it:

- _output_new_records_domain

- _input_pending_records_domain

Benefits:

- avoids repeated attachment subqueries during record selection

- keeps business behavior unchanged (records with/without file)

- improves query planning and execution time on large datasets

Query-level EXPLAIN ANALYZE on my tests:
Before (attachment subquery): 448.887 ms
After (has_exchange_file boolean): 240.222 ms
Gain: about 46% lower execution time, around 1.87x faster


edi_oca/models/edi_exchange_record.py
```diff
@@ -161,6 +165,29 @@ def _compute_exchange_filename(self):
             if not rec.exchange_filename:
                 rec.exchange_filename = rec.type_id._make_exchange_filename(rec)
 
+    @api.depends("exchange_file")
```
simahawk, line 168: wouldn't be easier to have an inverse on `exchange_file` to set the flag to True?
Then you need a migration script to set the flag by default.

simahawk (changes requested):

edi_oca/models/edi_exchange_record.py
santostelmo: Inverse on `exchange_file` ? This field is not computed
`exchange_file = fields.Binary(attachment=True, copy=False)`
what do you mean ?

edi_oca/models/edi_exchange_record.py
```diff
@@ -161,6 +165,29 @@ def _compute_exchange_filename(self):
...
+    def _compute_has_exchange_file(self):
+        if not self.ids:
+            for rec in self:
+                rec.has_exchange_file = False
```
SilvioC2C: Nitpicking:
```suggestion
            self.has_exchange_file = False
```
You can use the fields' multirecord setter

edi_oca/models/edi_exchange_record.py
SilvioC2C: @santostelmo yep, the fields' `inverse` attribute can be set on non-computed fields as well 😉

SilvioC2C (approved):
Overall LGTM; I don't have a strong opinion on whether making the field computed (as it is now) or having it set by an `exchange_file`'s `inverse` method (as suggested by @simahawk).
To make Codecov happy: please add automated tests using `odoo.tests.common.Form` and/or `self.env["edi.exchange.record"].new(vals)` to test what happens in the `if not self.ids` block.

Ricardoalso (commented):
I would implement the inverse approach suggested by @simahawk 

Thanks for the performance benchmark 🙏

edi_oca/models/edi_exchange_record.py
santostelmo: Thanks @simahawk for the suggestion. I added the inverse method

edi_oca/models/edi_exchange_record.py
santostelmo: Small unit test added for that

edi_oca/models/edi_exchange_record.py
```diff
@@ -161,6 +169,10 @@ def _compute_exchange_filename(self):
             if not rec.exchange_filename:
                 rec.exchange_filename = rec.type_id._make_exchange_filename(rec)
 
+    def _inverse_exchange_file(self):
+        for rec in self:
+            rec.has_exchange_file = True
```
simahawk: shouldn't it check if the file is set? What if I'm setting it to False to clean it?

edi_oca/tests/test_record.py
```diff
@@ -257,3 +257,16 @@ def test_related_records(self):
         self.assertEqual(len(record.related_record_ids), 2)
         self.assertEqual(record.related_record_ids[0].record, self.partner)
         self.assertEqual(record.related_record_ids[1].record, spain)
+
+    def test_compute_record_without_ids(self):
```
simahawk: what is this supposed to test?

simahawk (changes requested):

edi_oca/tests/test_record.py
santostelmo: Replaced by test_has_exchange_file

edi_oca/models/edi_exchange_record.py
santostelmo: Changed to ensure that writing `False` to `exchange_file` also clears `has_exchange_file`

SilvioC2C (approved):

Ricardoalso (approved):

